PR #14, opened 2018-12-14
Adds pytest

bwalsh:
This PR contains several new features:
* adds package module
* adds pytest
* adds requirements.txt,  requirements-dev.txt
* adds pre-commit checks
* adds README and CONTRIBUTING markdown

PR #15, opened 2018-12-18
Adds travis

ahwagner:
this builds off of #14 to add Continuous Integration with TravisCI


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
+dist: xenial
```
bwalsh: Not sure of impact in travis, but should we be using latest i.e cosmic?
http://releases.ubuntu.com/

ahwagner: Thanks for reviewing, @bwalsh!

PR #17, opened 2018-12-21
Adds & initializes DVC

bwalsh:
This PR contains one new feature:
* [dvc](http://dvc.org), backed by a [s3 bucket](https://s3-us-west-2.amazonaws.com/metakb-dvc/index.html)

The primary use cases are reproducibility and collaboration.  From dvc documentation.

```
Save and reproduce your experiments
At any time, fetch the full context about any experiment you or your team has run. DVC guarantees that all files and metrics will be consistent and in the right place to reproduce the experiment or use it as a baseline for a new iteration.

Version control models and data
DVC keeps metafiles in Git instead of Google Docs to describe and version control your data sets and models. DVC supports a variety of external storage types as a remote cache for large files.

Establish workflow for deployment & collaboration
DVC defines rules and processes for working effectively and consistently as a team. It serves as a protocol for collaboration, sharing results, and getting and running a finished model in a production environment.
```


ahwagner:
I'm not sure about the use of dvc + S3 for this project; we do have a git data server that we can use for data versioning, and it plays nicely as a submodule (e.g. the DGIdb project).

I agree with the need to version our data releases. I'm not familiar with dvc; is there any reason why it would be a better fit here than a data submodule?

.dvc/config
```diff
@@ -0,0 +1,4 @@
+['remote "metakb"']
+url = s3://metakb-dvc
```
ahwagner: Are there any size restrictions here? IIRC, last time we were butting up against free-tier availability. As we expand the captured data, will that be a problem for this bucket? If we go this route, should we instead use the WUSTL-hosted S3 account for the VICC project?

bwalsh:
Hi @ahwagner, happy new year.  Hope you had a good break.

There is a write up here that describes dvc.
https://dvc.org/doc/dvc-philosophy/related-technologies.  

For me the key benefits are:
* Independent of data backend (local file system, s3, gs, etc.).  Different instances can use the solution that fits for them.
* Provenance for free.  The commands and the hashes of the input/output files are maintained in git.  Reproducibility should be a non issue.
* Dependency management.  A user shouldn't need to know the intricacies of reproducing the workflow to a given stage.  

I'm sure there is more than one way to address this.  It would be good to address it early, as part of project setup.

.dvc/config
bwalsh: Yes, for sure we want to use an official WUSTL-hosted S3 account.

bwalsh:
@ahwagner  For sure we should have a discussion about what data solution works best.  (I was heads down and wanted to get the set of PRs done before the holiday break).  No commitment here to dvc, just wanted to get the conversation started.

bwalsh:
@ahwagner  - Pretty busy first week back from vacation.  Changes based on comments above committed.

PR #34, opened 2020-11-25
Issue 30: Loader/Harvester for CIViC

korikuzma:


jsstevenson (dismissed):
👍

jsstevenson (dismissed):
👍 

For the future -- it might be good to try to implement a retrieval method or store assertions under `self.assertions` so that the `get_assertions` method in the base class can be called successfully, but that might not be a particularly stable interface at present.

metakb/harvesters/civic.py
```diff
@@ -0,0 +1,377 @@
...
+    def harvest(self):
+        """Retrieve and store records from CIViC."""
+        try:
+            # civicpy.load_cache(on_stale='ignore')
+            evidence = self._harvest_evidence()
+            genes = self._harvest_genes()
+            variants = self._harvest_variants()
+            assertions = self._harvest_assertions()
+            self._create_json(evidence, genes, variants, assertions)
+            print('CIViC Harvester was successful.')
```
bwalsh, line 20: Use python logging instead of print  ?

metakb/harvesters/civic.py
```diff
@@ -0,0 +1,377 @@
...
+        assertions = civicpy.get_all_assertions()
+        assertions_list = list()
+
+        for assertion in assertions:
+            a = self._harvest_assertion(assertion)
+            assertions_list.append(a)
+        return assertions_list
+
+    def _harvest_gene(self, gene):
+        g = {
```
bwalsh, line 90: We could use a doc string on these `_harvest_<singular>()` methods.   The code looks fine, but I'm unclear as to the intent.

As I read the code on first pass, it seems they are defining a normalized structure for the entity.
In some cases, like this _harvest_gene, a new object is defined `g`.
In other cases, like _harvest_variant, the class variable is decorated with additional fields `v_extra`

metakb/harvesters/civic.py
```diff
@@ -0,0 +1,377 @@
...
+                }
+                for variant in gene.variants
+            ],
+            'aliases': gene.aliases,
+            # TODO: Add lifecycle_actions, sources
+            'type': gene.type
+        }
+
+        # Update evidence_items
+        for v in g['variants']:
```
bwalsh, line 111: Assume we are initializing an aggregator here.   Intent unclear, will other harvesters do the same?

.coveragerc
```diff
@@ -1,2 +1,7 @@
 [run]
-omit = setup.py
+source =
+   metakb/
+   tests/
+omit =
+   setup.py
+   docs/*
```
bwalsh, line 7: Suggest adding the following to ignore venv/ from coverage report
```     venv/*

```

analysis/civic/civic_harvester_example.py
```diff
@@ -0,0 +1,68 @@
+"""Create an example json file for CIViC Harvester."""
```
bwalsh, line 1: A README.md in this directory would be useful.  At first read, I thought it was creating test fixtures or notebook setup data.

metakb/models/__init__.py
```diff
@@ -0,0 +1 @@
+"""Data models for representing VICC records."""
```
bwalsh, line 1: Nice to see the models declare separately from the harvesters  (puzzled how they relate to the dictionaries in the `_harvest_<>(singular)`  methods)

tests/unit/harvesters/test_civic_assertions.py
```diff
@@ -0,0 +1,109 @@
...
+    """Create a list of genes."""
+    civicpy.load_cache(on_stale='ignore')
+    c = CIViC()
+    return c
+
+
+@pytest.fixture(scope='module')
+def aid40():
+    """Create a fixture for AID40 assertion."""
+    return {
```
bwalsh, line 18: Over time, these fixtures will drift as the source service changes.   Consider managing them outside of test source

bwalsh (changes requested):
Great work.   I've left a few comments below.
* I was able to quickly setup the environment
* Tests ran successfully the first time

I'm unclear regarding some of the mappings created by `_harvest_<singular>()` functions.

No problem if you want to  fix and merge or merge as is and follow up in separate PR.

.coveragerc
korikuzma: @bwalsh It looks like your suggestion isn't showing.

metakb/harvesters/civic.py
korikuzma: This is just for CIViC harvesting genes. CIViCpy stored a list of strings for accepted, rejected, and submitted counts. This was to emulate the way CIViC API displayed evidence items which is a dictionary where the key is accepted, rejected, or submitted and the value is an int of the counts.

requirements-dev.txt
```diff
@@ -1,8 +1,11 @@
...
 Sphinx==3.0.4
 flake8==3.8.2
+flake8-docstrings==1.5.0
 coverage<5
 pytest-cov==2.9.0
 coveralls==2.0.0
-sphinxjp.themes.basicstrap==0.5.0
\ No newline at end of file
+sphinxjp.themes.basicstrap==0.5.0
+civicpy~=1.1.2
```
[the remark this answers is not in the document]
korikuzma: It should be added. We use civicpy to get all variants, evidence, and genes. Without these lines, Travis CI failed.

metakb/harvesters/civic.py
korikuzma: A lot of the harvester code was redundant at first. All methods without 'harvest' in them (i.e. `_evidence_item`, `_variant`, `_assertion`) contain this redundant code that is used in all four CIViC records. Variants and Assertions contain extra attributes  (`v_extra` and `a_extra`) not found in the other records.

.gitignore
```diff
@@ -110,3 +110,14 @@ docs/images/*.ai
 # application level caches
 caches/
 
+# allow pytest override with pytest.ini
+pytest.ini
+
+# ide
+.idea/
+
```
jsstevenson, line 118: (self-interested suggestion from me)

```suggestion
# ide
.idea/
.vim/

```

metakb/harvesters/base.py
```diff
@@ -0,0 +1,25 @@
+"""A module for the Harvester base class"""
+
+
+class Harvester:
+    """A base class for content harvesters."""
+
```
jsstevenson, line 6: I wonder if we should have a default init() method to provide a guarantee for Alex's iter_assertion() method below? eg

```suggestion
def __init__(self):
    self.assertions = []

```

jsstevenson (approved):
👍

metakb/harvesters/base.py
korikuzma: Added in civic.py too

.coveragerc
bwalsh: Created a separate PR to add single line:

```   venv/*
```

bwalsh:
@korikuzma Looks good to me  +1

bwalsh (approved):
+1   nice work

MayLiu27 (approved):
looks good to me!

liux8 (approved):
Hi Kori,

have both you and James switched to Python 3.9?

Thanks,
Jeff

marianakhoul (approved):

korikuzma:
> Hi Kori,
> 
> have both you and James switched to Python 3.9?
> 
> Thanks,
> Jeff

Hey Jeff, 

James and I are using Python 3.8.

korikuzma:
> Hi Kori,
> 
> have both you and James switched to Python 3.9?
> 
> Thanks,
> Jeff

Hi @liux8 ,

James and I are using Python 3.8.

metakb/models/biomarkers.py
```diff
@@ -0,0 +1,29 @@
+"""Data models for Biomarkers and supporting concepts."""
+from dataclasses import dataclass
+from typing import List
+from enum import Enum
+
+
+class VariationType(Enum):
+    """A placeholder class that will be covered by the Variant Lexicon."""
+
+    PROTEIN_SUBSTITUTION = 1
```
ahwagner, line 10: This is subject to change. I think we may end up with a multi-attribute model for variation type (e.g. protein sequence, substitution type in lieu of protein_substitution)

ahwagner (approved):
Looks good, though there are a few comments that need addressing.

ahwagner (approved):

metakb/models/therapies.py
```diff
@@ -0,0 +1,33 @@
+"""Data models for Therapies and supporting concepts."""
+from dataclasses import dataclass
+from typing import List
+
+
+@dataclass
+class Treatment:
```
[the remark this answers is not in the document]
jsstevenson: Good question - we should ask [May 2020 Alex](https://github.com/cancervariants/metakb/commit/343e408b1644b27f4b16b1a8e60b6613f646fa2b) if he had specific data in mind!

Kidding aside, we're not wedded to it if you think the distinction is unnecessary -- we haven't seen any data that makes a difference.

PR #43, opened 2021-01-08
Issue 40

### LINKED ISSUE #40: Compute CIViC API delta

korikuzma:
Closes #40


metakb/deltas/civic.py
```diff
@@ -0,0 +1,139 @@
...
+            c.harvest(fn=fn)
+
+            with open(f"{PROJECT_ROOT}/data/civic/{fn}", 'r') as f:
+                new_civic = json.load(f)
+
+        delta = {
+            '_meta': {
+                'civicpy_version':
+                    pkg_resources.get_distribution("civicpy").version,
+                # TODO: Might change. Assuming we harvest when computing delta
```
jsstevenson, line 56: More broadly, we should think about what metadata to provide with harvested information (and how delta computation/update methods modify or augment it)

metakb/deltas/civic.py
```diff
@@ -0,0 +1,139 @@
...
+logger.setLevel(logging.DEBUG)
+
+
+class CIVICDelta:
+    """A class for computing CIViC deltas."""
+
+    def __init__(self, main_json, *args, **kwargs):
+        """Initialize the CIVICDelta class.
+
+        :param str main_json: The path to the main CIViC composite json file.
```
jsstevenson, line 20: I think some context about the role played by the `new_json` keyword could be informative here (eg what you said on Slack).

jsstevenson (approved):

metakb/deltas/civic.py
korikuzma: I agree. I don't believe CIViCpy includes metadata at the moment. I know @bwalsh mentioned keeping track of the tools and versions, so that's why I included lines 54-55. I believe the lines below (which are not shown) include a `date_harvested` field. What other fields were you thinking of including @jsstevenson ?

MayLiu27 (approved):

metakb/deltas/civic.py
korikuzma: Added

metakb/deltas/civic.py
korikuzma: I added `metakb_version`. I'm unsure if we need `civicpy_version`.

jsstevenson (approved):

MayLiu27 (approved):

PR #41, opened 2021-01-04
#31: Add MOAlmanac harvester

MayLiu27:
The Molecular Oncology Almanac Harvester is now ready (with example files). Please review it and leave any suggestions/comments. 
Thank you.

PR #44, opened 2021-01-08
#32: PMKB Harvester

### LINKED ISSUE #32: Write loader / harvester for PMKB

jsstevenson:
Close #32 

I included a short writeup about the structure of the input data [here](https://github.com/cancervariants/metakb/tree/issue-32/analysis/pmkb). You can see the full source data download [here](https://pmkb.weill.cornell.edu/therapies/downloadCSV.csv). Happy to hear input from Alex, Brian et al about how to form output data given this input structure.


metakb/harvesters/pmkb.py
```diff
@@ -0,0 +1,245 @@
...
+    def __init__(self):
+        """Set up harvester object"""
+        self.assertions = []
+
+    def harvest(self, data_dir: Path = PROJECT_ROOT / 'data' / 'pmkb') -> bool:
+        """Harvest PMKB source. Retrieve and store genes, variants, and
+        interpretations.
+
+        :param pathlib.Path data_dir: path to local PMKB data source directory
+        :return: `True` if successful, `False` otherwise.
```
korikuzma, line 31: Does this ever return False?

jsstevenson: Good catch, I meant to return to this -- IMO down the line we might want to rethink whether returning true/false from this method is an optimal way of managing success and failure (not saying it's not, just seems like it's worth talking out). I'll add a basic exception catch in the meantime -- but I think at that point we could also go back and make some rules for how we want to catch generic exceptions (there are apparently small differences between a catch-all `except`, `except Exception`, and `except BaseException`, see eg https://realpython.com/the-most-diabolical-python-antipattern/)

korikuzma (approved):
Looks like everything was captured! 👍

metakb/harvesters/pmkb.py
```diff
@@ -0,0 +1,249 @@
...
+                'type': 'gene',
+                'name': gene,
+                'variants': [
+                    {
+                        'name': name,
+                        'evidence_count': count,
+                    }
+                    for name, count in var_counts.iteritems()
+                ]
+            })
```
MayLiu27, line 126: it seems PMKB doesn't provide additional gene-specific data, same as MOA. Alex suggested to drop the gene entry, we might wanna consider the same for PMKB? 
check: https://github.com/cancervariants/metakb/pull/41#discussion_r555359648

MayLiu27 (commented):
Great work! Looks good! Everything works correctly on my local machine. 
For the gene entry, please refer to Alex's comment.

ahwagner:
Per discussion at last MetaKB meeting, we are going to pause this PR until we get a response from PMKB re: API access and concept identifiers.

@jsstevenson please switch this over to a draft PR until we have clarity on that.

metakb/harvesters/pmkb.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :param string fn: file name of composite JSON document
+        :return: bool True if operation is successful, False otherwise
+        """
+        try:
+            self.pmkb_dir = PROJECT_ROOT / 'data' / 'pmkb'
+            self._check_files()
+            variants = self._get_all_variants()
+            statements = self._get_all_interpretations(variants)
+            self._create_json(statements, variants, fn)
```
korikuzma, line 30:
```suggestion
            self._create_json(statements, variants, fn)
            return True
```

metakb/harvesters/pmkb.py
```diff
@@ -0,0 +1,200 @@
...
+                continue
+            variant_id = variant['PMKB URL'].split('/')[-1]
+
+            variant_object = {
+                "name": name,
+                "gene": {
+                    "name": variant['Gene'],
+                },
+                "id": variant_id,
+                "origin": variant['Germline/Somatic'].lower(),
```
korikuzma, line 95: Not a huge deal, but since we're just harvesting, should we be altering the data?

metakb/transform/pmkb.py
```diff
@@ -0,0 +1,454 @@
...
+        assoc_with = []
+        cosmic_id = variant.get('cosmic_id')
+        if cosmic_id:
+            assoc_with.append(f'{schemas.XrefSystem.COSMIC.value}:{cosmic_id}')
+        ens_id = variant.get('ensembl_id')
+        if ens_id:
+            assoc_with.append(f'{schemas.XrefSystem.ENSEMBL.value}:{ens_id}')
+        if assoc_with:
+            vod['extensions'] = schemas.Extension(
+                name="associated_with",
```
korikuzma, line 314: I might be wrong on this
```suggestion
                name="xrefs",
```

jsstevenson: I'm gonna ask Alex about this to confirm but I think xrefs makes sense here

jsstevenson:
FYI -- a recent update of therapy-normalization (I think 0.2.19?) is a hard requirement and you'll have to rebuild your local therapy database (`python3 -m therapy.cli --normalizer="hemonc wikidata ncit chemidplus drugbank rxnorm" --update_merged`). The good news is that as of the latest version this should be much, much quicker, but let me know if you hit any issues with it.

metakb/transform/pmkb.py
```diff
@@ -0,0 +1,480 @@
...
+        """Get Descriptor objects given interpretation and associated data.
+
+        :param Dict interpretation: PMKB interpretation
+        :param Dict variants: Keys are variant labels and values are all
+            harvested variant objects
+        :param Dict gene_index: lookup gene ID from symbol
+        :param Dict disease_index: lookup disease ID from label
+        :return: Tuple containing Lists of therapy, disease, gene, and
+            variant descriptors (we expect each to be len == 1)
+        """
```
korikuzma, line 152: `gene_index` and `disease_index` not used 
```suggestion
    def _get_descriptors(self, interpretation, variants):
        """Get Descriptor objects given interpretation and associated data.

        :param Dict interpretation: PMKB interpretation
        :param Dict variants: Keys are variant labels and values are all
            harvested variant objects
        :return: Tuple containing Lists of therapy, disease, gene, and
            variant descriptors (we expect each to be len == 1)
        """
```

metakb/transform/pmkb.py
```diff
@@ -0,0 +1,480 @@
...
+        if therapy in invalid_keys:
+            return []
+
+        vod = self.transformed['therapy_descriptors'].get(therapy)
+        if vod:
+            return [vod]
+
+        response = self.vicc_normalizers.normalize_therapy([therapy])
+        if not response or not response[0] or response[0]['match_type'] == 0:
+            print(therapy)
```
korikuzma, line 198:
```suggestion
```

metakb/transform/pmkb.py
```diff
@@ -0,0 +1,480 @@
...
+        vod = schemas.VariationDescriptor(
+            id=f"pmkb.variant:{variant['id']}",
+            type="VariationDescriptor",
+            label=label,
+            value_id=response.value_id,
+            value=response.value,
+            gene_context=gene_id,
+            molecule_context=response.molecule_context,
+            structural_type=response.structural_type,
+            ref_allele_seq=response.ref_allele_seq,
```
korikuzma, line 323: Might be wrong because I still get confused on this... I don't think we should be including normalizer responses in fields other than value object fields. If I am wrong, then let me know and we should make these additions to CIViC and MOA.

korikuzma (changes requested):
Great job with this massive PR! Just some small comments and suggestions that might be useful.

metakb/harvesters/pmkb.py
```diff
@@ -0,0 +1,237 @@
...
+            if name in variants_out:
+                logger.error(f"Multiple records for variant: {name}")
+                continue
+            variant_id = variant['PMKB URL'].split('/')[-1]
+
+            variant_object = {
+                "name": name,
+                "gene": {
+                    "name": variant['Gene'],
+                },
```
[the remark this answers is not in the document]
jsstevenson: Only if you don't like needlessly complex things!

metakb/transform/pmkb.py
jsstevenson: No, I think you have it correct -- the Value comes from the normalizer, but the descriptor refers to the way that the Value is described by the resource.

MayLiu27 (dismissed):
Looks good to me. Great work!

jsstevenson:
@ahwagner @bwalsh 

Any feedback on the material in `analysis/pmkb` would be helpful -- I wrote a few notes on roadblocks/where we stand [here](https://github.com/cancervariants/metakb/tree/issue-32/analysis/pmkb#notes-about-data), but that folder also includes sample input and harvest/transform output.

jsstevenson:
I'm going to close this for now -- I think we can make some progress given concurrent updates to the rest of MetaKB but we can return to it later

ahwagner:
Was going to make this a draft PR instead, but don't have the permissions... so re-closing.

jsstevenson:
Seems like open -> draft PR is not an option so I will... have repeated the above for nothing

korikuzma:
hehe

PR #52, opened 2021-01-28
Delta model for CIViC and MOAlmanac

korikuzma:


MayLiu27 (dismissed):
Looks Great! I changed the comment comment under __init__ from CIViCDelta to Delta, minor change.

metakb/deltas/delta.py
```diff
@@ -1,119 +1,124 @@
...
-    def __init__(self, main_json, *args, **kwargs):
-        """Initialize the CIVICDelta class.
+    def __init__(self, main_json, src, *args, **kwargs):
+        """Initialize the Delta class.
 
-        :param str main_json: The path to the main CIViC composite json file.
+        :param str main_json: The path to the main composite json file.
+        :param str src: The source to compute the delta on
         """
+        assert src.lower() in ['civic', 'pmkb', 'moa']
```
ahwagner, line 22: I wonder if we can codify the source list somewhere in our schema as an enum, and draw from that enum to generate this list and other places such a list might be used? I think it would be a nice, self-documenting way to improve code maintainability and query validation.

metakb/deltas/delta.py
```diff
@@ -1,119 +1,124 @@
-"""A module for computing CIViC deltas."""
+"""A module for computing deltas."""
 from metakb import PROJECT_ROOT
 import json
 import logging
 from jsondiff import diff
 from datetime import date
-import pkg_resources
-from metakb.harvesters import CIViC
+from metakb.harvesters import CIViC, MOAlmanac
```
ahwagner, line 8: Adding a dict here to reference classes by string can be used to reduce complexity in lines 50-55.

```suggestion
HARVESTER_CLASS = {
    'civic': CIViC,
    'moa': MOAlmanac
}
```

metakb/deltas/delta.py
```diff
@@ -1,119 +1,124 @@
...
         else:
             # Want to create updated harvester file
-            fn = f"civic_harvester_{current_date}.json"
-            c = CIViC()
-            c.harvest(fn=fn)
+            fn = f"{self._src}_harvester_{current_date}.json"
+            if self._src == 'civic':
+                CIViC().harvest(fn=fn)
+            elif self._src == 'moa':
+                MOAlmanac().harvest(fn=fn)
```
ahwagner, line 54: With respect to earlier comment defining the HARVESTER_CLASS dict:

```suggestion
            fn = f"{self._src}_harvester_{current_date}.json"
            HARVESTER_CLASS[self._src](fn=fn)
```

metakb/deltas/delta.py
ahwagner: This could even be automatically generated through class inspection, and/or defined and imported from the harvesters package __init__.py.

ahwagner (commented):
This looks great. A few minor places where we may consider some changes are commented on below.

With this change, may also want to consider ditching the `deltas` package (which only contains `delta.py` and an `__init__.py`, and making `delta.py` a top-level module.

ahwagner (approved):

PR #53, opened 2021-02-16
Issue 49

### LINKED ISSUE #49: harvester testing
The CIViC and MOAlmanac harvesters currently test fixtures against live queries to the resources. This causes test failures as resources update content, even if the code is behaving as expected.

We should work around this. One strategy is to build [mock interfaces](https://docs.pytest.org/en/stable/monkeypatch.html?highlight=mocking#monkeypatching-returned-objects-building-mock-classes) and [prevent request sessions](https://docs.pytest.org/en/stable/monkeypatch.html?highlight=mocking#global-patch-example-preventing-requests-from-remote-operations).

korikuzma:
Closes #49


Pipfile
```diff
@@ -14,6 +14,7 @@ requests = "*"
 numpy = "*"
 jsondiff = "*"
 requests-cache = "*"
+mock = "*"
```
jsstevenson, line 17: Should probably be under dev, right? (presumably true of pytest and the other testing packages as well)

jsstevenson (approved):
👍

jsstevenson:
Oh, one other thought -- I think we can remove the harvester base class from coverage analysis, given that it sort of distorts the percentage and doesn't need any testing itself

Pipfile
korikuzma: You're right. I thought the other packages were fixed, but they were only fixed in the `requirements.txt` and `requirements-dev.txt`

jsstevenson (approved):
👍

PR #55, opened 2021-03-15
Issue 47

korikuzma:
Travis will continue to fail due to AWS credentials, which are changed every 12 hours, for the vicc normalizers. We're currently looking into this for the normalizers and will update MetaKB once it's resolved.


metakb/transform/civic.py
```diff
@@ -0,0 +1,488 @@
...
+        return [disease_descriptor]
+
+    def _get_therapy_descriptors(self, drugs):
+        """Return a list of Therapy Descriptors.
+        :param dict drugs: Drugs for a given evidence_item
+        :return: A list of Therapy Descriptors
+        """
+        therapies = list()
+        if len(drugs) != 1:
+            return therapies
```
jsstevenson, line 415: is there any way to combine this check with the stuff on line 93-94 or do they cover different angles?

jsstevenson (approved):

metakb/transform/civic.py
```diff
@@ -0,0 +1,488 @@
...
+
+        if disease_norm_id.startswith('ncit:'):
+            disease_descriptor = schemas.ValueObjectDescriptor(
+                id=f"civic:did{disease['id']}",
+                type="DiseaseDescriptor",
+                label=display_name,
+                value=schemas.Disease(disease_id=disease_norm_id),
+            ).dict()
+        else:
+            # TODO: Should we accept other disease_ids other than NCIt?
```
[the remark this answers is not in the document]
korikuzma: Ahh okay, that's good to know! I've checked my logs and no civic disease records were logged from the evidence_items we're able to use. 
When I made this comment, I was wondering what we'd do if our normalizer couldn't find a NCIt match. Would it be acceptable to use one of the lower ranking matches (Mondo, OMIM, Oncotree, DO)  in the disease value object? CIViC always stores the DOID, so I didn't know if that would be acceptable.

metakb/transform/civic.py
```diff
@@ -0,0 +1,499 @@
...
+            so_id = variant['variant_types'][0]['so_id']
+            if so_id in ['SO:0001583', 'SO:0001818']:
+                structural_type = 'SO:0001060'
+                molecule_context = 'protein'
+            elif so_id in ['SO:0001886', 'SO:0001576', 'SO:0001889']:
+                structural_type = 'SO:0001060'
+                molecule_context = 'transcript'
+            else:
+                # TODO: Genomic
+                pass
```
korikuzma, line 279: Is there a better way to find the `molecule_context`? I was thinking of going through SO and try to get a list of possible values, but SO has so many concepts and if it doesn't say `protein`, `transcript`, or `genomic` then I don't know what it classifies as. 
For CIViC, it appears that the `structural_type` is always `SO:0001060` (I'll set this in line 265 and remove lines 272 and 275).

metakb/transform/civic.py
```diff
@@ -0,0 +1,499 @@
...
+                           f" Variant Normalization normalize.")
+            return []
+
+        variation_descriptor = schemas.VariationDescriptor(
+            id=f"civic:vid{variant['id']}",
+            label=variant['name'],
+            description=variant['description'],
+            value_id=normalized_resp.value_id,
+            value=normalized_resp.value,
+            gene_context=f"civic:gid{gene['id']}",
```
korikuzma, line 305: `gene_context` is probably wrong. I think we'd put the `coordinates.representative_transcript` (if this value exists) from a [variant record](https://civicdb.org/api/variants/38), i.e. `gene_context="ensembl:ENST00000269571.5"`?

bwalsh (approved):
Looks good.  I like the `Test CIViC Transformation to common data model` integration test

MayLiu27 (commented):
Looks good to me! Great work! 
One other questions is what about the part we discussed appending the kb own resources value to `extensions`?

metakb/transform/civic.py
```diff
@@ -0,0 +1,718 @@
...
+        disease_norm_id = \
+            disease_norm_resp['value_object_descriptor']['value']['disease_id']
+
+        if disease_norm_id.startswith('ncit:'):
+            disease_descriptor = schemas.ValueObjectDescriptor(
+                id=f"civic:did{disease['id']}",
+                type="DiseaseDescriptor",
+                label=display_name,
+                value=schemas.Disease(disease_id=disease_norm_id),
+            ).dict()
```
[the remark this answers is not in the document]
korikuzma: I believe that these values should come from the source itself.

metakb/transform/civic.py
```diff
@@ -0,0 +1,718 @@
...
+            id=f"civic:vid{variant['id']}",
+            label=variant['name'],
+            description=variant['description'] if variant['description'] else None,  # noqa: E501
+            value_id=variant_norm_resp.value_id,
+            value=variant_norm_resp.value,
+            gene_context=f"civic:gid{gene['id']}",
+            molecule_context='protein',
+            structural_type='SO:0001060',
+            ref_allele_seq=re.split(r'\d+', variant['name'])[0],
+            expressions=hgvs_exprs,
```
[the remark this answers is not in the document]
korikuzma: You're right for a `Value Object Descriptor`. But here, I'm using a [Variation Descriptor](https://vrsatile.readthedocs.io/en/latest/value_object_descriptor/vod_index.html#variation-descriptor) which contains additional fields.

metakb/transform/civic.py
```diff
@@ -0,0 +1,765 @@
...
+                continue
+
+            gene_descriptors = self._get_gene_descriptors(
+                self._get_record(evidence['gene_id'], genes))
+
+            documents = self._get_evidence_document(evidence['source'],
+                                                    propositions_documents_ix)
+
+            assertion_methods = [schemas.AssertionMethod(
+                id='assertion_method:001',
```
jsstevenson, line 111: I'm guessing we will a) need some kind of method to create unique ID numbers and/or b) namespace this somehow? So that eg we don't have conflicts between assertion methods tied to different sources, etc

korikuzma: Good catch. I can add something to `schemas.py` to store these constants for now.

korikuzma: Added in [this commit.](https://github.com/cancervariants/metakb/commit/1be7dcdf94dd5d95817a9dfbec22408945a7c544)

jsstevenson: Okay, I misremembered what the Assertion Method class is used for, but I agree that declaring some constants is the way to go forward (especially for people like me who commit typos at a rate of once per minute)

ahwagner (changes requested):
A few updates to schemas.py. Overall this is looking good, let me know when these changes have been addressed and I'll do a full review.

Also, I may have a few additional comments on Evidence and Assertions–possibly today.

metakb/schemas.py
```diff
@@ -0,0 +1,274 @@
...
+    disease_context: str  # vicc:Disease
+    therapy: str  # Therapy value object
+
+
+class AssertionMethodID(IntEnum):
+    """Create constraints for AssertionMethod ids for harvested sources."""
+
+    CIVIC_EID_SOP = 1
+    CIVIC_AID_AMP_ASCO_CAP = 2
+    CIVIC_AID_ACMG = 3
```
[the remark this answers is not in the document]
korikuzma: This is for the `AssertionMethod.id` . Originally, I hardcoded the `id` for CIViC `assertion_methods`. @jsstevenson pointed out that this would cause duplicates once MOA and other sources are added. As of now, we're using these constants across sources we harvest to ensure uniqueness. Let me know if you think there's a better way to do this!

jsstevenson (approved):

analysis/civic/examples/transform/civic:aid6.json
```diff
@@ -0,0 +1,237 @@
...
+                "support_evidence:497",
+                "support_evidence:498",
+                "support_evidence:499"
+            ],
+            "type": "Statement"
+        }
+    ],
+    "propositions": [
+        {
+            "_id": "proposition:124",
```
ahwagner, line 27:
```suggestion
            "id": "proposition:124",
```
one more change: let's get away from the VRS `_id` usage. We have a model that can support `value_id` separate from `value`, so we can do away with `_id` entirely; no need to mirror in propositions.

analysis/civic/examples/transform/civic:aid6.json
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    "syntax": "hgvs:genomic",
+                    "value": "NC_000007.13:g.55259515T>G",
+                    "version": null,
+                    "type": "Expression"
+                }
+            ],
+            "ref_allele_seq": "L",
+            "gene_context": "civic:gid19"
+        }
```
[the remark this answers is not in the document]
korikuzma: Should I have not put `civic:vid33.rep` in an extension as [shown here](https://github.com/cancervariants/metakb/blob/issue-47/analysis/civic/examples/transform/civic:aid6.json#L72)?

analysis/civic/examples/transform/civic:aid6.json
```diff
@@ -0,0 +1,237 @@
...
+            "method": "method:002",
+            "support_evidence": [
+                "support_evidence:493",
+                "support_evidence:494",
+                "support_evidence:495",
+                "support_evidence:496",
+                "support_evidence:497",
+                "support_evidence:498",
+                "support_evidence:499"
+            ],
```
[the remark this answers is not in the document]
korikuzma: Changed `support_evidence` to `supported_by`. For CIViC EIDs, I used `civic:eid###` rather than `civic:eid.###` to stay consistent with what we have been using. If you'd rather `civic:eid.###`, I can make that change.

analysis/civic/examples/transform/civic:aid6.json
```diff
@@ -0,0 +1,237 @@
...
+            "url": "https://pubmed.ncbi.nlm.nih.gov/27993330/",
+            "version": {
+                "year": 2017,
+                "month": 1,
+                "day": null
+            },
+            "reference": "Li MM, Datto M, Duncavage EJ, et al."
+        }
+    ],
+    "support_evidence": [
```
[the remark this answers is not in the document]
korikuzma: Renamed `support_evidence` to `documents` and added CIViC EIDs to the list of `statements`.

metakb/transform/civic.py
```diff
@@ -0,0 +1,837 @@
...
+        """Get a CIViC record by ID.
+
+        :param str record_id: The ID of the record we are searching for
+        :param dict records: A dict of records for a given CIViC record type
+        """
+        for r in records:
+            if r['id'] == record_id:
+                return r
+
+    def _add_to_list(self, eid, key, list_name):
```
MayLiu27, line 804: I don't see this method has been used, could be removed if not.

MayLiu27 (dismissed):
Looks good to me. Great work! Just one method indicated below could be removed if not used.

metakb/transform/civic.py
korikuzma: Good catch! Removing now.

PR #67, opened 2021-04-08
Issue 47 refactor

korikuzma:
- Changes output to lists (similar to harvest response)
- Value ID does not need to only be NCIt
- Run all possible queries through normalizer until highest match is found


MayLiu27 (approved):
Looks good!

PR #68, opened 2021-04-09
Issue 54 refactor

MayLiu27:
Changed the format to be similar as harvester, and changed the id selection from normalizer. Renamed some variable names to be consistent with what civic.py uses. 

Only disease_Descriptor used the highest_match checking method since that's the only case it has multiple queries to try. 
I'm updating the MOA API responses in issue-54

### DIFF
--- a/metakb/transform/moa.py
+++ b/metakb/transform/moa.py
@@ -21,7 +21,7 @@ class MOATransform:
     def __init__(self,
                  file_path=f"{PROJECT_ROOT}/data/moa/moa_harvester.json"):
         """
-        Initialize MOATransform class
+        Initialize VICC normalizers and class attributes
 
         :param: The file path to the harvested json to transform
         """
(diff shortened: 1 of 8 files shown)


metakb/transform/moa.py
```diff
@@ -32,27 +32,48 @@ def __init__(self,
...
-        moa_dir = PROJECT_ROOT / 'data' / 'moa' / 'transform'
+    def _create_json(self,
+                     moa_dir=PROJECT_ROOT / 'data' / 'moa' / 'transform',
+                     fn='moa_cdm.json'):
+        """Create a composite JSON for the transformed MOA data.
+
+        :param path moa_dir: The moa transform data directory
+        :param str fn: The file name for the transformed data
+        """
+        print(moa_dir)
```
korikuzma, line 57:
```suggestion
```

korikuzma (approved):

PR #75, opened 2021-04-12
Maps mounted data dependencies to normalizer packages

bwalsh:
This PR adds a docker file that includes a script `discover_normalizer_paths.sh`  that discovers the pipenv paths and maps mounted directories to package paths.

```
     - ./data/therapy:/app/therapy/data          
     - ./data/gene:/app/gene/data
     - ./data/disease:/app/disease/data
```


richardhj:
```Dockerfile
RUN chmod +x discover_normalizer_paths.sh
```
needs to be added, otherwise, you may run into a "command not found in pipenv path" error.

korikuzma:
Going to close since this is an old PR.

PR #79, opened 2021-04-20
Issue 74

### LINKED ISSUE #74: Search Endpoint Detail Flag
We currently reference `variation_descriptor`, `therapy_descriptor`,  `disease_descriptor`, `method`, and `supported_by` by ID in the search endpoint. We should add a `detail` flag to allow users to see all of these values.

korikuzma:
Closes #74

### DIFF
--- a/tests/unit/test_search.py
+++ b/tests/unit/test_search.py
@@ -533,29 +533,49 @@ def assert_keys_for_detail_true(response_keys, response, is_evidence=True):
             assert len(response[field]) > 1
 
 
-def assert_same_ids(response):
-    """Assert that IDs match in response items."""
-    statement = response['statements'][0]
-    proposition = response['propositions'][0]
-    vd = response['variation_descriptors'][0]
-    gd = response['gene_descriptors'][0]
-    td = response['therapy_descriptors'][0]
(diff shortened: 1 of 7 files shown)


tests/unit/test_search.py
```diff
@@ -501,6 +874,48 @@ def test_multiple_parameters(query_handler):
     assert len(response['matches']['propositions']) > 1
 
 
+def test_civic_detail_flag(query_handler, civic_eid2997, eid2997_proposition,
```
jsstevenson, line 877: It'd probably have a really ugly method signature, but would it be possible/helpful to write a helper function to perform the logic for these detail flag tests and have the CIViC and MOA test cases call it? Might make adding more sources later a bit easier

jsstevenson (approved):

tests/unit/test_search.py
korikuzma: Yup, I'll add it!

korikuzma: Added in this commit https://github.com/cancervariants/metakb/pull/79/commits/45bff62b7513f4d67b010ae82caa2ec758661785

MayLiu27 (approved):
Great work!

PR #89, opened 2021-04-30
Fix staging

### LINKED ISSUE #88: Search By ID Endpoint Errors on Invalid Node IDs
When searching on an ID that does not exist in the database:
> return query.search_by_id(id)
Apr 30 15:56:24 ip-10-130-5-114 web: File "/var/app/current/metakb/query.py", line 780, in search_by_id
Apr 30 15:56:24 ip-10-130-5-114 web: concept_name = quote(node_id.split(":", 1)[1])
Apr 30 15:56:24 ip-10-130-5-114 web: IndexError: list index out of range

korikuzma:
Closes #88

### DIFF
--- a/metakb/query.py
+++ b/metakb/query.py
@@ -775,31 +775,35 @@ class QueryHandler:
         :param str node_id: node_id
         :return: A dictionary containing the node content
         """
-        node_id = node_id.strip()
-        if '%' not in node_id:
-            concept_name = quote(node_id.split(":", 1)[1])
-            node_id = node_id.split(":", 1)[0] + ":" + concept_name
+        valid_node_id = None
         response = {
-            'query': None,
+            'query': node_id,
             'warnings': []
         }
 
         if not node_id:
             response['warnings'].append("No parameters were entered.")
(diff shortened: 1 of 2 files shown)


metakb/query.py
```diff
@@ -775,20 +775,23 @@ def search_by_id(self, node_id):
...
         }
 
         if not node_id:
             response['warnings'].append("No parameters were entered.")
+        elif node_id.strip() == '':
+            response['warnings'].append("Cannot enter empty string.")
         else:
-            valid_node_id = None
-            response['query'] = node_id
+            node_id = node_id.strip()
```
MayLiu27, line 789: maybe add another `node_id = node_id.replace(" ", "")`? When I was testing some queries, I always accidentally typed a space right after `":"`

MayLiu27 (approved):

metakb/query.py
korikuzma: Hmmm... That's more user error. I don't think we should do this since MOA could have an ID that is a label with spaces in it.

PR #122, opened 2021-11-03
Issue 69

### LINKED ISSUE #69: pypi package
Add MetaKB package to pypi

korikuzma:
Closes #69 

Notes:
- move data dir inside metakb app
    - Create `harvester`, `transform`, and `delta` directories for each source
- updated requirements (mainly updating normalizer packages)
- updated version (let me know if this should be something different. I don't think we should be at 2.0.0-alpha yet..)
- Added new endpoint (`search/statements`) for getting nested statements 
- Added type hints, but all of this will be addressed in #123 
- Removed unused / old models directory


korikuzma:
I think this is essentially done. Waiting on [this](https://github.com/cancervariants/variation-normalization/pull/179) PR from variation normalizer and will then run again to make sure everything looks okay.

korikuzma:
Also want to address [this](https://github.com/cancervariants/variation-normalization/issues/99) issue in variation normalizer because it's causing one of our search tests to fail

metakb/main.py
```diff
@@ -63,6 +64,23 @@ def search(variation: Optional[str] = Query(None, description=v_description),
                         detail)
 
 
+@app.get('/api/v2/search/statements',
+         summary=search_summary,
+         response_description=search_response_description,
+         response_model=SearchStatementsService,
+         description=search_description,
```
jsstevenson, line 71: We should (either now or later) update descriptions to distinguish this endpoint from `/search`

setup.cfg
```diff
@@ -1,2 +1,52 @@
...
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.8
+
+[options]
+packages = find:
+python_requires = >=3.8
+zip_safe = False
+
+install_requires =
```
jsstevenson, line 26: Do we want to put the dependencies that stuff like elastic beanstalk requires? I think it was this group:

```
uvloop = "*"
websockets = "*"
httptools = "*"
```

metakb/query.py
```diff
@@ -307,27 +331,336 @@ def search(self, variation='', disease='', therapy='', gene='',
...
+                proposition = \
+                    TherapeuticResponseProposition(**proposition)
+            elif proposition_type == PropositionType.PROGNOSTIC:
+                proposition = PrognosticProposition(**proposition)
+            elif proposition_type == PropositionType.DIAGNOSTIC:
+                proposition = DiagnosticProposition(**proposition)
+            else:
+                # TODO Check
+                raise Exception(f"{proposition_type} is not a valid "
+                                f"proposition type")
```
jsstevenson, line 636: maybe this should be a ValueError or something more specific?

metakb/query.py
```diff
@@ -307,27 +331,336 @@ def search(self, variation='', disease='', therapy='', gene='',
...
+
+        session = self.driver.session()
+        statement_nodes = list()
+        proposition_nodes = session.read_transaction(
+            self._get_propositions, normalized_therapy,
+            normalized_variation, normalized_disease, normalized_gene,
+            valid_statement_id
+        )
+
+        proposition_cache = dict()
```
jsstevenson, line 449: It might be worth looking at abstracting away the method calls that populate each cache and wrap them with lru_cache or something similar, rather than manually building/destroying caches in each query

metakb/query.py
```diff
@@ -307,27 +331,336 @@ def search(self, variation='', disease='', therapy='', gene='',
             response['documents'] = None
 
         session.close()
-        return SearchService(**response).dict(exclude_none=True)
+        return SearchService(**response).dict(by_alias=True, exclude_none=True)
+
+    def search_by_id(self, node_id: str) -> SearchIDService:
```
jsstevenson, line 336: this function returns a Dict, not a dataclass

jsstevenson (approved):
nothing super mandatory, maybe just some issues to create and handle later

metakb/query.py
korikuzma: Lol left the todo comment to change this but forgot about it

setup.cfg
korikuzma: No, these are only used for production deployment purposes

metakb/main.py
korikuzma: completely forgot about this. good catch. i also need to update schemas

metakb/main.py
```diff
@@ -63,6 +64,23 @@ def search(variation: Optional[str] = Query(None, description=v_description),
...
+         response_model_exclude_none=True)
+def get_statements(
+        variation: Optional[str] = Query(None, description=v_description),
+        disease: Optional[str] = Query(None, description=d_description),
+        therapy: Optional[str] = Query(None, description=t_description),
+        gene: Optional[str] = Query(None, description=g_description),
+        statement_id: Optional[str] = Query(None, description=s_description)):
+    """Return nested statements for queried concepts"""
+    return query.search_statements(
+        variation, disease, therapy, gene, statement_id)
```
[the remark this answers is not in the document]
korikuzma: I think I'll just change the method parameters in the QueryHandler class, because the `Default value:` on FastAPI is weird

korikuzma:
Realized I missed some EB things + updated ga4gh.vrsatile.pydantic locally but did not reflect these changes in Pipfile / requirements

jsstevenson (dismissed):

MayLiu27 (approved):
Nice work. Only thing is in the analysis - harvest, should change import CIViC, MOAlmanac to CIViCHarvester and MOAlmanacHarvester.

MayLiu27 (approved):
Looks Great!

PR #141, opened 2021-12-22
Issue 127

### LINKED ISSUE #127: Create versioned data files
We should version harvesters and transformed CDM JSON files similar to what we do in our normalizers. For example, `civic_harvester20211207` and `civic_cdm20211208`

jsstevenson:
Fix #127 

* Removes the option to set filename in method parameters -- this wasn't something we were using (and doesn't seem to be something we do in any of our other related projects)


korikuzma (changes requested):
I think we should keep the option to set the filename so we can run older versions. It's not implemented in our normalizers, but I think it is something we should look towards adding. Other than that it looks good!

metakb/transform/base.py
```diff
@@ -11,24 +15,49 @@
...
+        :param Optional[Path] harvest_path: path to harvest JSON file. If not
+        provided, will attempt to open file with current date at default
+        location.
+        :return: Dict containing Lists of entries for each object type
+        """
+        if not harvest_path:
+            today = dt.strftime(dt.today(), DATE_FMT)
+            default_fname = f"{self.name}_harvester_{today}.json"
+            default_path = self.data_dir / "harvester" / default_fname
+            if not default_path.exists():
```
korikuzma, line 57: I think we should check this when `harvest_path` is set as well

korikuzma (changes requested):
Should also update analysis to use `MOAHarvester` + use versioned files

analysis/civic/examples/transform/civic_cdm_20211227.json
```diff
@@ -5,7 +5,7 @@
             "description": "A phase III clinical trial (NCT00949650) found that median progression free survival among patients with exon 19 deletions or L858R EGFR mutations (n = 308) was 13.6 months for afatinib and 6.9 months for chemotherapy (HR, 0.47; 95% CI, 0.34 to 0.65; P = 0.001).",
             "direction": "supports",
             "evidence_level": "civic.evidence_level:B",
-            "proposition": "proposition:127",
+            "proposition": "proposition:5k7StVvnNBf6qxTgXgZBCSaiyWScvS3Q",
```
korikuzma, line 8: I assume this will be removed in the proposition / document issue?

korikuzma (approved):

analysis/civic/examples/transform/civic_cdm_20211227.json
jsstevenson: ??? hold the phone

jsstevenson: I think this was happening because the CIViC examples generator just uses the first 5 assertions, and this one ^ no longer comes up in that list (so it was left over from my work on another branch). I cleared out my working folders and reran everything and it should be a bit less weird now

korikuzma (approved):

PR #149, opened 2022-01-17
Enable parameter entry on swaggerui by default

jsstevenson:


korikuzma (approved):
So glad they finally made this feature

Issue #180, opened 2022-07-19
Update source schemas for metaschema update


korikuzma:
Based off of [this](https://github.com/ga4gh/va-spec/tree/cvc-variation-descriptors) branch.

- [x] CIViC
- [x] MOAlmanac

By Aug 10, 2022

korikuzma:
Fixed in `metaschema-update` branch. This will eventually be merged into `staging` + `main`.

Issue #193, opened 2022-09-27
Update queries to reflect gk pilot update


korikuzma:

korikuzma:
@jsstevenson let me know your thoughts on these

- I would like to remove /api/v2/search . In this endpoint, values are referenced by ID (compared to the /api/v2/search/statements which nests the object). I don't believe this endpoint is used (at least not in VarCat). If we want to add it back, I think we should make a separate issue for it. I'm kind of going with the YAGNI
- I would like to change what is returned in /api/v2/search/{id} . The current implementation returns the populated CDM. I am proposing to only return node properties and labels such as below. While this endpoint is not used currently, it is very easy to maintain. 

  ```json
  {
    "query": "civic.eid:2997",
    "warnings": [],
    "node": {
      "predicate": "predictsSensitivityTo",
      "alleleOrigin": "somatic",
      "description": "Afatinib, an irreversible inhibitor of the ErbB family of tyrosine kinases has been approved in the US for the first-line treatment of patients with metastatic non-small-cell lung cancer (NSCLC) who have tumours with EGFR exon 19 deletions or exon 21 (L858R) substitution mutations as detected by a US FDA-approved test",
      "id": "civic.eid:2997",
      "direction": "supports"
    },
    "node_labels": [
      "VariantTherapeuticResponseStudy",
      "Study"
    ]
  }
  ```

korikuzma:
I'm going to make this an epic and split out the endpoints into new issues

PR #196, opened 2022-10-13
feat: initial work for metaschema updates

### LINKED ISSUE #181: Update civicpy cache
https://github.com/cancervariants/metakb/blob/main/metakb/harvesters/civic.py#L25

We should set `on_stale=update`. I think this is why we're missing some aliases in BRAF V600E in civic. The API was updated, so we'll have to checkout [this](https://github.com/griffithlab/civicpy/tree/civic_v2) branch for civicpy.

### LINKED ISSUE #189: Add support for CombinationTherapeuticCollection
Addresses issue in #128

### LINKED ISSUE #190: Add support for TherapeuticSubstituteGroup
Addresses #128

korikuzma:
Notes:
- Sorry for the very large PR. This PR addresses initial work for #180 . Looked at this for too long, so there's probably things I've missed / could clean up so feedback is much appreciated. 
- Added flake8 (I added since I was going to change most of the files anyway)
- Removed `detect-aws-credentials` from pre-commit. I'm not sure why this was added, but it's not necessary
- Includes initial analysis for CIViC + MOA metaschema updates
- CIViC Harvester
  -  civicpy was updated to reflect v2 API changes. Some fields we renamed / removed in this update.
  - Added `update_cache` param to `harvester` method in order to update civicpy cache (closes #181)
- I changed the way we were calculating proposition digests to follow how vrs-python does it. We can revert this if you want.
- For transforming to CDM, we only do Therapeutic for now.  
- Adds support for Combination Therapeutic Collection (closes #189)
- Adds support for Substitute Therapeutic Collection (closes #190)


metakb/service_schema.py
```diff
@@ -0,0 +1,259 @@
...
+
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any],
+                         model: Type['NestedStatementResponse']) -> None:
+            """Configure OpenAPI schema"""
+            if 'title' in schema.keys():
+                schema.pop('title', None)
+            for prop in schema.get('properties', {}).values():
+                prop.pop('title', None)
+            # TODO: Update
```
jsstevenson, line 61: ? is this pending an Alex contribution

there are also some other TODO: updates on this page as well fwiw

metakb/service_schema.py
```diff
@@ -0,0 +1,259 @@
...
+
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any],
+                         model: Type['Matches']) -> None:
+            """Configure OpenAPI schema"""
+            if 'title' in schema.keys():
+                schema.pop('title', None)
+            for prop in schema.get('properties', {}).values():
+                prop.pop('title', None)
+            schema['example'] = {
```
jsstevenson, line 140: fyi, there are some examples of single quotes on this page even though it looks like you went through and double-quoted everything elsewhere (maybe double check that a linter is active?)

metakb/transform/base.py
```diff
@@ -7,19 +7,121 @@
...
 
 
 class Transform:
     """A base class for transforming harvester data."""
 
+    _methods: List[Method] = [
+        Method(
+            id=MethodId.CIVIC_EID_SOP.value,
+            is_reported_in=Document(
+                xrefs=["pmid:31779674"],
```
jsstevenson, line 28: the DOI for this is `10.1186/s13073-019-0687-x`

metakb/transform/base.py
```diff
@@ -7,19 +7,121 @@
...
+                label="Danos AM, Krysiak K, Barnell EK, et al., 2019, Genome Medicine",
+                title="Standard operating procedure for curation and clinical "
+                      "interpretation of variants in cancer",
+            ).dict(exclude_none=True),
+            label="CIViC Curation SOP (2019)").dict(exclude_none=True),
+        Method(
+            id=MethodId.MOA_ASSERTION_BIORXIV.value,
+            is_reported_in=Document(
+                xrefs=["doi:10.1038/s43018-021-00243-3"],
```
jsstevenson, line 37: the PMID for this is `35121878`

metakb/transform/base.py
```diff
@@ -128,6 +254,17 @@ def _get_document_id(**parameters) -> str:
...
         return f"document:{sha512t24u(blob=blob)}"
 
+    @staticmethod
+    def _get_digest_for_str_lists(l: List[str]) -> str:  # noqa: E741
+        """Create digest for a list of strings
+
+        :param List[str] l: List of strings to get digest for
+        :return: Digest
+        """
+        l.sort()
```
jsstevenson, line 264: low impact but I think avoiding "l" as a variable name is probably a good practice (maybe you could use "strings" as a name instead)

jsstevenson (changes requested):

metakb/transform/base.py
korikuzma: Ugh I get these warnings and ignore them. Naming is hard.

metakb/service_schema.py
korikuzma: Oh, I forgot to note that the service_schema won't be addressed til #193 . This was just a copy/paste into a new file.

metakb/service_schema.py
korikuzma: This commit was probably done before I added the flake8 pre-commit hook. It will also be addressed in #193

metakb/transform/base.py
korikuzma: Renaming to `str_list`

jsstevenson (approved):

PR #245, opened 2023-11-10
feat!: update transformers to use vrs 2-alpha and gk-pilot schemas

### LINKED ISSUE #230: Transformers use VRS 2.0-alpha for updated gk pilot data
Needed for plenary

re: @ahwagner 

> We now have complete and passing schemas and examples for the GK-Pilot v2 work, including our updated categorical variation classes (from VRSATILE), VRS 2.0a, and the updated statement classes for VA. You can find this work in the following branches:
> * vrs -> 2.0-alpha
> * vrsatile -> gk-pilot
> * va-spec -> gk-pilot
>
> The schemas in these repos are all built on GKS Metaschema processor 0.2.1  and JSON Schema 2020-12.
> Complete example statements live in the va-spec repo that validate from the [VarStatement schema](https://github.com/ga4gh/va-spec/blob/gk-pilot/schema/varstatement.json). These examples are available in the fixtures directory:
> [ClinVar SCV ex1](https://github.com/ga4gh/va-spec/blob/gk-pilot/tests/fixtures/VA-ClinVar-SCV-Example-001.yaml)
> [ClinVar SCV ex2](https://github.com/ga4gh/va-spec/blob/gk-pilot/tests/fixtures/VA-ClinVar-SCV-Example-002.yaml)
> [CIViC EID ex1](https://github.com/ga4gh/va-spec/blob/gk-pilot/tests/fixtures/tr-statement-inline.yaml)

korikuzma:
close #230 

Related tests:

```shell
python3 -m pytest tests/unit/transform/test_civic_transform_therapeutic.py tests/unit/transform/test_moa_transform.py
```

Things done in this PR:
* This PR only focuses on updated CIViC (evidence) and MOA (assertions) for therapeutic response with single therapy. Other study/statement types will be added in separate issues. The db, query, and cli modules will also be updated in separate issues. They do not work at the moment. 
* Updates ga4gh/normalizer dependencies to point to VRS 2-alpha versions
  * Removes ga4gh.vrsatile.pydantic since ga4gh.vrs/ga4gh.core has the models 
  * Updates to Pydantic V2
* Updated schemas
  * For now leaving in MetaKB. Maybe in future, we can create a new package for Pydantic schemas (if there is a need)
  * I was having trouble using `datamodel-code-generator` to generate the pydantic classes, so I manually created. We can create a new issue (similar to vrs-python) to get `datamodel-code-generator` to work how we want. 
  * Annotation and Variation Statement schemas can be found [here](https://github.com/ga4gh/va-spec/tree/5e6ee0a3a6847200684534667b96431ce844c9f5/schema)
  * Categorical schemas can be found [here](https://github.com/ga4gh/vrsatile/tree/348958fe5e9c35d79e7daa0c4732ffcc47dece2f/schema)
  * Used private classes (class name starts with underscore) for schemas in source but not the final yaml/json. 

Notes:
* schemas
  * I wasn't sure how to handle `extends: {field}` in `schemas/variation_statement` schemas. I just created a `_StatementBase` class with the fields that do not get extended and used the new field name in the associated schema. 
  * There is no `RecordMetadata` schema even though it is referenced in the va-spec schemas
  * I think `VariantStatement` should have variant required (as shown below). The source does not include this, but the final yaml/json has the variant field required for child classes. 
  
    ```yaml
    heritableRequired:
        - variant
    ```


korikuzma:
@jsstevenson I think this is mainly ready. I've been reviewing in pieces, but want to do one more final review before I mark this ready for review.

metakb/transform/civic.py
```diff
@@ -25,553 +68,519 @@ def __init__(self,
...
-            variant_query = f"{variant['entrez_name']} {variant_name}"
-            hgvs_exprs = self._get_hgvs_expr(variant)
+    @staticmethod
+    def _is_supported_variant_query(variant_name: str, variant_id: int) -> bool:
+        """Determine if a variant name is supported by the variation-normalizer.
+        This is used to skip normalization on variants that the variation-normalizer
+        is known not to support
+
+        :param variant_name: Variant name in CIViC
+        :param variant_id: CIViC Variant ID
```
jsstevenson, line 397: return description (easy)

metakb/transform/civic.py
```diff
@@ -25,553 +68,519 @@ def __init__(self,
...
-                       if v is not None}
-            ).dict(exclude_none=True),
-            Extension(
-                name='civic_actionability_score',
-                value=variant['civic_actionability_score']
-            ).dict(exclude_none=True)
-        ]
+            # Create VRS Variation object
+            params = vrs_variation.model_dump(exclude_none=True)
+            params["id"] = vrs_variation.id
```
jsstevenson, line 470: is this line needed? shouldn't `params` already have this field?

jsstevenson (changes requested):

metakb/schemas/annotation.py
```diff
@@ -0,0 +1,160 @@
...
+        """Check that date is YYYY-MM-DD format"""
+        if v:
+            valid_format = "%Y-%m-%d"
+
+            try:
+                datetime.strptime(v, valid_format).strftime(valid_format)
+            except ValueError:
+                raise ValueError("`date` must use YYYY-MM-DD format")
+
+        return v
```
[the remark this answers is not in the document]
korikuzma: Thank you. I was not doing the correct search!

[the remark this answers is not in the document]
korikuzma: So date should be a string and not a datetime.date

metakb/transform/civic.py
korikuzma: Right, thanks!

jsstevenson (approved):

PR #247, opened 2023-11-16
refactor!: fix class name casing

### LINKED ISSUE #246: Fix class name casing
For example, I think `VICCNormalizers` should be renamed to `ViccNormalizers`

korikuzma:
close #246 

@jsstevenson wasn't sure about CLI? Should I just change to `Cli`?


jsstevenson:
> @jsstevenson wasn't sure about CLI? Should I just change to Cli?

I don't think it matters bc it probably doesn't need to be a class -- seems like it's all static methods besides the main CLI entry point (don't need to change right now)

jsstevenson (approved):

PR #276, opened 2024-01-03
tests: update civic + moa test data

korikuzma:
```shell 
python3 -m pytest tests/unit/harvesters/ tests/unit/transform/ -vv
```

* Updates civic + moa test data from 20240103
* Updated moa ids that changed


jsstevenson (approved):
> Updated moa ids that changed

👎

jsstevenson:
there should be an emoji for 💯 but it's -100

Issue #274, opened 2024-01-02
Update `/search` endpoint for gk-pilot schemas


korikuzma:

korikuzma:
I think we can probably close this. It wasn't an actual requirement and no one has asked for it. We can re-open if it is needed.

Issue #272, opened 2024-01-02
Update `search/{id}` endpoint for gk-pilot schemas


korikuzma:

korikuzma:
This is not a requirement at the moment. We can re-open if we do need to add it back

Issue #270, opened 2023-12-21
Create enums for database node labels and relationships


korikuzma:
It might be beneficial to do this to use in both database and query modules

PR #266, opened 2023-12-20
feat!: update database to reflect schema changes

korikuzma:
Original work for this was done in #264 . This was incomplete. So starting a fresh PR

@jsstevenson This could use a new set of eyes.

`python3 -m pytest tests/unit/harvesters/ tests/unit/transform/ tests/unit/database/`

* Update neo4j to 5.x driver
* Update database nodes and relationships
  * Dropped Variation Group. We can add back in new issue if we want to retain this
* Changed fixture scope in `tests/conftest.py` to `session`
* Cleaned up getting secret value

PR #279, opened 2024-01-04
feat!: update `/search/studies` endpoint

### LINKED ISSUE #273: Create `/search/studies` endpoint for gk-pilot schemas
This will replace the current `/search/statements` endpoint

korikuzma:
close #273 . I will open once #266 is merged

@jsstevenson again, fresh eyes would be helpful for this 😄 

Notes:
* This PR updates the previously named `/search/statements` endpoint. Will address #309 at a later time. This PR is to kind of mirror what was done previously, so I'd like to address "major" response changes in separate issues if possible.
* There are some places where I should use enums. I will do this in #270 or a separate issue. I opened an issue in VRS-Python to create str enums
* For input variation, I decided to query based on a `definingContext` or a `members` match
* The way that identifiers are computed in VRS-Python made it so that our FastAPI endpoints only return the minimal information needed (ga4gh keys) to create a digest. See more in #305 . I haven't had time to dive much deeper into this issue

Changes:
* Renamed `/search/statements` to `/search/studies` and updated to use the latest schema
* Temporarily removes other endpoints `/search` and `/search/{id}` which will be added in #274 and #272


korikuzma:
I think I will wait to open until @bwalsh and @ahwagner give feedback on dev instance.

After initial talk with @ahwagner , I should move `variant.definingContext.expressions` into `variant.members`. For CIViC variants, the genomic hgvs expression should already be there. 

Screenshot 2024-01-08 at 10 03 41 AM

korikuzma:
The last commit added support for providing genomic variation query. At the moment, only CIViC studies will be returned. Once #283 is complete, MOA assertions should also be returned

tests/conftest.py
```diff
@@ -2153,50 +2255,94 @@ def moa_source44():
...
+                                        {
+                                            "name": f"{normalizer_data_type}_normalizer_id",  # noqa: E501
+                                            "type": "Extension",
+                                            "value": ext["value"]["normalized_id"]
+                                        }
+                                    )
+                                else:
+                                    new_extensions.append(ext)
+                            if extensions:
+                                v["extensions"] = new_extensions
```
korikuzma, line 2306: I don't love this. This was done because CDM expose more fields (this was needed for varcat) than exposed in the search endpoint. Because I wanted to get this out fast, I did not expose this additional information in the endpoint. I think this would require a plugin such as [APOC](https://neo4j.com/docs/apoc/current/) to help us access fields in a dict in cypher queries but didn't know if it would be difficult to setup on the ec2 instance.

korikuzma:
> I think I will wait to open until @bwalsh and @ahwagner give feedback on dev instance.
> 
> After initial talk with @ahwagner , I should move `variant.definingContext.expressions` into `variant.members`. For CIViC variants, the genomic hgvs expression should already be there.
> 
> Screenshot 2024-01-08 at 10 03 41 AM

Made new issue for this #309

korikuzma:
> Something is happening with the `civic.eid:2880` evidence item and associated queries such as `NC_000013.10:g.32890600G>T` and `BRCA2 M1I`. We are getting a `No studies found with the provided query parameters.` in the warnings, but the `study_ids` field has `civic.eid:2880` in it. I will look into this later.
> 
> Using 20240103

#310 fixes this

metakb/version.py
```diff
@@ -1,4 +1,2 @@
 """MetaKB version"""
-# REQ: EACH TIME VERSION IS UPDATED, MUST ALSO UPDATE LAST_UPDATED
-__version__ = "1.1.0"
-LAST_UPDATED = "2022-11-08"
+__version__ = "2.0.0-dev0"
```
jsstevenson, line 2: WOW i was wondering about this

korikuzma: This was an old thing that was needed for VarCat, but it's no longer being used

jsstevenson (commented):
👍 looks good

metakb/query.py
```diff
@@ -70,1214 +103,550 @@ def get_normalized_disease(self, disease: str,
...
-                sb_id = sb.get("id")
-                try:
-                    if sb_id in document_cache:
-                        document = document_cache[sb_id]
+                    if node_type == "CombinationTherapy":
+                        node["components"] = self._get_therapeutic_agents(
+                            tx, node["id"], TherapeuticProcedureType.COMBINATION,
+                            TherapeuticRelation.HAS_COMPONENTS
+                        )
+                        params["therapeutic"] = core_models.TherapeuticProcedure(**node)
```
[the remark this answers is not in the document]
korikuzma: Yes, good catch!

korikuzma:
@jsstevenson applied your changes. I also don't love the ordering of methods. I can re-order after you review latest changes

jsstevenson (dismissed):
👍

jsstevenson (approved):
yeah seems reasonable

PR #284, opened 2024-01-19
style: use ruff

jsstevenson:
It'd be better for everyone's health if this waits until other PRs go through (managing conflicts will be a nightmare otherwise). But was going through the repo and noticed a few things, so I wanted to just get PRs and issues up.


jsstevenson:
Closing, since it'll need to be redone w/ new ruff configs anyway

PR #294, opened 2024-03-22
style: use ruff

jsstevenson:
Need to rebase/update branch once `pyproject.toml` PR goes through. Also need to actually implement fixes on all rule violations.


jsstevenson:
as with #292 need to redo PR base to merge into `epic-231`

PR #306, opened 2024-03-27
chore: remove unused delta class

### LINKED ISSUE #277: Remove unused Delta class
We should consider removing this class. Originally, we were going to use this for Kafka but we not have used this class in 3 years. I have not run it since then, so I do not know if it still works. If down the line we need this, we can add it back with cleaned up code.

korikuzma:
close #277


jsstevenson (approved):
👍 easy

PR #310, opened 2024-03-27
fix: creating location parameterized query string

korikuzma:
Original issue from: https://github.com/cancervariants/metakb/pull/279#issuecomment-1923806985

* start can be 0, so needed to do null check


jsstevenson (approved):
mhm good thought

PR #316, opened 2024-03-29
cicd: adjust installation checks

### LINKED ISSUE #314: Tweak Python version distributions on install check action
We currently have an action that checks whether the Pipfile installs correctly, since that's what beanstalk uses. However, we run this check on 3.8-3.12, even though Beanstalk presumably only runs on a specific version. 

Maybe it would make more sense to have actions that check pip/setuptools based installation across every version (since this is how we recommend non-devs install it) and then a separate action focused specifically on Pipenv and the Python version used on elastic beanstalk.

jsstevenson:
close #314

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -15,7 +15,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: python3 -m pip install ".[tests,dev,notebook]"
+        run: python3 -m pip install ".[tests,dev,notebooks]"
 
   pipenv_install:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yaml
```diff
@@ -15,6 +14,21 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install dependencies
+        run: python3 -m pip install ".[tests,dev,notebook]"
```
korikuzma, line 18:
```suggestion
        run: python3 -m pip install ".[tests,dev,notebooks]"
```

korikuzma (changes requested):

.github/workflows/checks.yaml
jsstevenson: every time!

korikuzma (changes requested):
@jsstevenson change the base branch to ‘staging’. Only elastic beanstalk apps follow this workflow atm

korikuzma (approved):

PR #323, opened 2024-04-02
fix: add support for civic ash source type

### LINKED ISSUE #322: Add support for civic ASH source in transformer
using s3://vicc-metakb/civicpy/cache/cache-20240402.pkl.zip, civic transformer failed document validation since we don't support this source

korikuzma:
close #322

### DIFF
--- a/src/metakb/transform/civic.py
+++ b/src/metakb/transform/civic.py
@@ -1,7 +1,7 @@
 """A module for to transform CIViC."""
 import logging
 import re
-from enum import StrEnum
+from enum import Enum
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


src/metakb/transform/civic.py
```diff
@@ -898,7 +902,11 @@ def _add_eid_document(self, source: Dict) -> Optional[Document]:
             if document not in self.documents:
                 self.documents.append(document)
         else:
-            logger.debug("%s not in schemas.SourcePrefix", source_type)
+            logger.debug(
```
jsstevenson, line 905: Might be worth making this a higher warning level? I'd think of "debug" as more appropriate for cases like "this is a very granular detail about what is happening because something mysterious is breaking and I'm tracking the pipeline step by step", whereas this would be something that I'd want to know about even if the pipeline runs successfully

jsstevenson (dismissed):

jsstevenson (approved):

PR #326, opened 2024-04-02
cicd: fix elastic beanstalk deployment

korikuzma:
Elastic beanstalk deployment was failing due to this

* Procfile command updated for new src layout


jsstevenson:
Ohhhh dang yeah should've caught that

jsstevenson (approved):

PR #332, opened 2024-05-09
feat!: sharpen up type usage

### LINKED ISSUE #285: Add type annotations
@korikuzma, didn't see an existing issue for this, apologies if I missed it

A lot of this code is from pretty far back in the day, and the type checker is flagging a lot of inconsistencies/missing annotations. It'd be good to get this fixed where possible.

In #284 I just totally removed checking annotation rules because it should probably be done separately. Once finished, that checker should go back on.

jsstevenson:
close #285 

* Clean up types and type annotations where possible. Not exhaustive -- there's some stuff that probably needs refactoring to make progress on -- but made some good progress, such that I think we can say the "types" issue is closed enough.
* **maybe important**: by virtue of where json model dumps were happening, this introduced a small change to the order of the keys in some dicts that were getting sha512'd, which meant the ID changed during testing. I set JSON serialization to first sort keys in order so that this doesn't matter in the future, but this does entail a one-time change in digest ID. if that's bad, I can revert and try to figure out a workaround.
* Also changed how stuff from GKS/VRS models are imported so that the code is a little more succinct. I think `models.Variation` style imports were a relic from a Python-JSONschema-objects custom and we don't need to do it anymore.


src/metakb/database.py
```diff
@@ -16,7 +16,7 @@
 
 
 def _create_parameterized_query(
-    entity: Dict, params: Tuple[str], entity_param_prefix: str = ""
+    entity: Dict, params: Tuple[str, ...], entity_param_prefix: str = ""
```
korikuzma, line 19: Did not know about this type hints!

src/metakb/transform/base.py
```diff
@@ -284,13 +293,13 @@ def _get_digest_for_str_lists(str_list: List[str]) -> str:
         :return: Digest
         """
         str_list.sort()
-        blob = json.dumps(str_list, separators=(",", ":")).encode("ascii")
+        blob = json.dumps(str_list, separators=(",", ":"), sort_keys=True).encode(
+            "ascii"
+        )
```
korikuzma, line 298: When you squash + merge, I think this should be included in the commit message and not just descr since this is where the breaking change occurred

src/metakb/transform/civic.py
```diff
@@ -95,6 +104,16 @@ class SourcePrefix(str, Enum):
...
 
+    # Cache for normalized concepts. The key is the concept type and value is a
+    # dictionary of mappings from CIViC concept (key) to transformed concept (value)
+    able_to_normalize: ClassVar[Dict[str, Dict]] = {
+        "variations": {},  # will store _VariationCache data
+        "molecular_profiles": {},
+        "diseases": {},
+        "therapeutics": {},
+        "genes": {},
+    }
```
korikuzma, line 115: I'm not sure about having this as a class var. This will change based on the `harvester_path` and I think the changes should be isolated for each instance

src/metakb/transform/moa.py
```diff
@@ -32,6 +41,16 @@
...
 
+    # Cache for normalized concepts. The key is the concept type and value is a
+    # dictionary of mappings from MOA concept (key) to transformed concept (value)
+    able_to_normalize: ClassVar[Dict[str, Dict]] = {
+        "variations": {},
+        "diseases": {},
+        "therapeutics": {},
+        "genes": {},
+        "documents": {},
+    }
```
korikuzma, line 52: Same comment as above

src/metakb/transform/moa.py
```diff
@@ -148,14 +157,14 @@ async def _add_variant_therapeutic_response_studies(
                     # skipping HORMONE and CHEMOTHERAPY for now
                     continue
 
-                therapies = [tn.strip() for tn in therapy_name.split("+")]
+                therapies = [{"label": tn.strip()} for tn in therapy_name.split("+")]
```
korikuzma, line 160: I think I may be missing the reason for using a dict, can you explain this change?

src/metakb/transform/base.py
jsstevenson: Yeah I've been copying/pasting PR descriptions into the squash commits lately, since it also adds them to the local git history and doesn't just live on github

src/metakb/transform/moa.py
jsstevenson: yeah -- in MOA, `get_therapeutic_agent` had broken the interface [defined in the abstract class](https://github.com/cancervariants/metakb/blob/836f47d63bebbea31708735e62ca80d4a468caf9/src/metakb/transform/base.py#L302) by making the argument a string. I assume this is because in CIViC, drugs are more richly described, but they're just a single name in MOA. I think the simplest way to make a generalization that's true for both is to just have the MOA drugs just be simple dicts with a single key/value.

src/metakb/transform/base.py
korikuzma: We can change the default squash and merge settings to use PR title / description so less copy/pasta

src/metakb/transform/moa.py
korikuzma: Sounds good! Thanks

src/metakb/transform/civic.py
jsstevenson: I'd moved them out here because some methods in the `Base` class presuppose this field's existence -- I settled for defining an empty dict in `__init__()` and then filling it in when the child classes initialize

korikuzma (approved):

PR #333, opened 2024-05-10
build: raise support to require python >=3.10

jsstevenson:
Wanted to move ahead with this while letting https://github.com/GenomicMedLab/software-templates/pull/29 simmer a bit, since it does affect a bunch of the type syntax in #332


jsstevenson:
Oh, right, this will require doing the new type annotation here and now. Ok.

PR #334, opened 2024-05-10
cicd: add issue templates, stalebot, and branch action

korikuzma:


jsstevenson (approved):
glad I could sneak one final issue in before this

PR #338, opened 2024-05-10
refactor!: clean up harvest method

korikuzma:
* `harvest` no longer saves data to a file, instead a pydantic class is returned
* Harvester classes no longer have instance variables storing record data
* Added new method `save_harvested_data_to_file` to save harvested data to a JSON file


src/metakb/harvesters/base.py
```diff
@@ -11,14 +12,23 @@
...
-        :rtype: bool
+        :param harvested_filepath: Path to the JSON file where the harvested data will
+            be stored. If not provided, will use the default path of
+            ``<APP_ROOT>/data/<src_name>/harvester/<src_name>_harvester_YYYYMMDD.json``
+        """
+        harvested_data = self.get_harvester_data()
+        self.create_json(harvested_data, harvested_filepath=harvested_filepath)
+
+    @abstractmethod
+    def get_harvester_data(self) -> dict[str, list[dict]]:
```
korikuzma, line 26: Naming is hard. This or something else?

jsstevenson: I think this is a hard thing to name because the `harvest()` method skips a step in terms of logical data processing. You'd expect something like

```
acquire data -> transform it -> return it -> (maybe) save it
```

but instead `harvest` acquires, transforms, and saves it, and then `get_harvester_data` returns the state of the data from right before `harvest` saved it.

anyway I think this is fine the way it is -- maybe rename `harvest()` to `harvest_to_file` so that there's a clearer distinction between what these methods do?

src/metakb/harvesters/civic.py
```diff
@@ -35,50 +35,24 @@ def __init__(
...
+        variants = self.harvest_variants()
+        molecular_profiles = self.harvest_molecular_profiles()
+        assertions = self.harvest_assertions()
+        return {
+            "evidence": evidence,
+            "genes": genes,
+            "variants": variants,
+            "molecular_profiles": molecular_profiles,
+            "assertions": assertions,
+        }
```
jsstevenson, line 55: out of scope for this PR IMO, but one thing that makes some of the higher metakb methods so messy is how much stuff is built into deeply nested dicts. I wonder if it'd be better to rely on namedtuples/pydantic classes to enforce structure on objects passed around internally, instead of just waiting until the web API layer to put them through the pydantic classes.

jsstevenson (approved):

src/metakb/harvesters/civic.py
korikuzma: Does it make sense for us to use Pydantic classes for harvested unstructured data?

jsstevenson: I don't think you can for the stuff you're ingested, since its structure could just be whatever... on the other hand, the harvester data object is technically a dict but it's a dict with specific field names (and we wouldn't expect a consumer to alter those field names) so you could just make it a namedtuple or pydantic class so that those field names are known in advance

korikuzma: Okay, that makes more sense. I thought you were suggesting pydantic models for the unstructured data. That's an easy change, so I can just include it in this PR

korikuzma:
@jsstevenson I'm going to do more cleanup

jsstevenson (approved):

PR #339, opened 2024-05-10
feat: prevent usage of aws db instances in cli

### LINKED ISSUE #331: Make sure CLI can't overwrite cloud normalizer DBs

korikuzma:
close #331

* if aws db environment variables are set, cli exits

### DIFF
--- a/src/metakb/cli.py
+++ b/src/metakb/cli.py
@@ -14,8 +14,18 @@ import boto3
 from boto3.exceptions import ResourceLoadException
 from botocore.config import Config
 from disease.cli import update_db as update_normalizer_disease_db
+from disease.database.database import AWS_ENV_VAR_NAME as DISEASE_AWS_ENV_VAR_NAME
+from disease.database.database import (
+    SKIP_AWS_DB_ENV_NAME as DISEASE_SKIP_AWS_DB_ENV_NAME,
+)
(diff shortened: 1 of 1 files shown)


src/metakb/cli.py
```diff
@@ -215,11 +225,24 @@ def _help_msg(msg: str = "") -> None:
...
+        "Gene": (GENE_SKIP_AWS_DB_ENV_NAME, GENE_AWS_ENV_VAR_NAME),
+    }
+
     for name, update_normalizer_db_fn in [
         ("Disease", update_normalizer_disease_db),
         ("Therapy", update_normalizer_therapy_db),
         ("Gene", update_normalizer_gene_db),
     ]:
+        skip_aws_db_env_name, aws_env_var_name = aws_env_map[name]
+        if environ.get(skip_aws_db_env_name) == "true" and aws_env_var_name in environ:
```
jsstevenson, line 240: does `skip_aws_db_env_name` matter? From the message below, it seems like we want to prohibit updating the cloud DBs from the metakb CLI regardless.

jsstevenson (changes requested):
(potentially requesting changes depending on clarification)

src/metakb/cli.py
korikuzma: Ya it probably is easiest to remove it than manually select y/n

jsstevenson (approved):

PR #344, opened 2024-05-14
feat!: break out CLI for more specific commands/more reasonable defaults

### LINKED ISSUE #210: Consider updating CLI arguments
I think we should consider a few things with the CLI:
- [ ] When using the default arguments (`python3 -m metakb.cli`), we run _all_ source harvesters and _all_ source transformations. It might be good to include a `--update_sources` argument (can be named something else) which default is all the sources, but users can indicate if they would only like to update a subset of the sources. 
- [ ] It might also be useful to allow users to run transformations on previously harvested data. So we would allow users to provide the harvester name + harvester path to use for the transformation.

jsstevenson:
close #210 

Miscellaneous quality of life improvements and feature additions to CLI:

* Add `metakb` as a console command.
* Break out CLI subcommands:
    * `metakb update` to run a complete harvest/transform/load step
    * `metakb check-normalizers` and `metakb update-normalizers` to check and force refreshing of normalizer data. This supports a simple workflow like `metakb check-normalizers || metakb load-normalizers` to load if unavailable, rather than requiring the user to force normalizer reload while loading the MetaKB graph.
    * `metakb harvest` to just perform harvest of source(s)
    * `metakb transform` to just perform transform of source(s), or `metakb transform-file` to transform a specific harvested file
    * `metakb load-cdm` to skip harvest/transform and directly load a CDM file, either from local (default location), a specific file, or from S3
    * `metakb clear-graph` to wipe the graph. No other CLI command will wipe the graph. I thought about calling it when `update` is used without any source qualifiers, but it seemed a little odd to include additional behavior such that `metakb update <source> && metakb update <other source>` is different from `metakb update`. Also thought about including it as an option flag in some other commands, but at that point, you can just do `metakb clear-graph && <other command>`.
    * Previously, we had one command that could be used to do a lot of things. IMO it's cleaner and more intuitive (and easier to program/maintain) to have subcommands with specific purposes instead of one big one. It does mean you will often have to chain commands together but that's pretty normal.
* Support selection of specific sources for commands where it makes sense. Pass them as arguments. Otherwise default to all sources.
* Support output directory option (`--output_directory`, `-o`) where it makes sense. Unfortunately, most of these commands all produce `n` output files so I don't think there's a simple way to specify the name of the output file.
* Collapse the credentials CLI option to a single `username:password` option, since you need to provide both at once. (Not sure why Neo4j *requires* a password).
* When updating normalizers, keep going to the next one if one of them fails.
* In general, give precedence to explicit args/options over env vars. Consequently, pass normalizer DB URL as an arg to normalizers rather than injecting it via env var to the transform step, which required a small refactor.
* Refactor a few normalizer management functions out of the CLI module. In general, I think it's good to separate reusable functions out of CLI modules, and just have `cli.py` act as gateways/interfaces to them. There's probably a little bit more of this that we could do but nothing else stuck out to me.
* Various changes to make console printouts a bit prettier/more organized.


korikuzma:
@jsstevenson did you see this old issue #210?

jsstevenson:
one possible additional todo: add -o to things that produce file outputs to specify output location

src/metakb/cli.py
```diff
@@ -14,178 +14,384 @@
...
+        if aws_env_var_name in environ:
+            msg = (
+                f"Updating the {name.value} AWS database from the MetaKB CLI is "
+                f"prohibited. Unset the environment variable:`{aws_env_var_name}` to "
+                "proceed."
+            )
+            _logger.error(msg)
+            click.echo(msg)
+            success = False
+            continue
```
jsstevenson, line 161: Once we're fully set up to update normalizers in the cloud, I wonder if there's some configuration we could add to the Dynamo table templates that only allow writes to come from the identity that's granted to the cloud update services. It'd be nice if we didn't have to write all of these checks into the code itself.

src/metakb/cli.py
```diff
@@ -14,178 +14,384 @@
...
+
+    To prepare normalizers if unavailable, invalidate cached data, then load MetaKB data:
+
+    \b
+        $ metakb check-normalizers || metakb load-normalizers
+        $ metakb update --update_source_caches
+    """  # noqa: D301
+
+
+_normalizer_db_url_description = "URL endpoint of normalizer database. If not given, the individual normalizers will revert to their own defaults."
```
jsstevenson, line 75: Since Dynamo takes a single URL but Postgres requires different libpq-style URLs that include normalizer-specific database names, I don't think it's really feasible to support postgres connections this way yet. Thought about doing something that could take a base postgres URL and add database names on top but it's probably not worth the work right now.

jsstevenson:
@korikuzma having some second thoughts about what commands/combinations of commands should call `graph.clear()`

korikuzma:
> @korikuzma having some second thoughts about what commands/combinations of commands should call `graph.clear()`

@jsstevenson I haven't looked at the changes, but what if we just separate this out and place it on the user to decide and make a note in the documentation?

src/metakb/schemas/app.py
```diff
@@ -1,9 +1,32 @@
...
+
+    def __repr__(self) -> str:
+        """Print as simple string rather than enum wrapper, e.g. 'civic' instead of
+        <NormalizerName.CIVIC: 'civic'>.
+
+        Makes Click error messages prettier.
+
+        :return: formatted enum value
+        """
+        return f"'{self.value}'"
```
jsstevenson, line 32: Bigger explanation of this:

For the most part, when you give enum values to `click.Choice`, it'll format them correctly. For example:

```
[ cli-refactor ⚙ .venv] ~/code/metakb % metakb update --help
Usage: metakb update [OPTIONS] [[civic|moa]]...
```

However, if you pass an invalid arg, it prints them using `__repr__`, so it's ugly:

```
Error: Invalid value for '[[civic|moa]]...': 'sdflkd' is not one of <NormalizerName.CIVIC: 'civic'>, <NormalizerName.MOA: 'moa'>.
```

I don't think there are any major consequences to overwriting `__repr__` but when debugging you have to be a little careful because it'll print without the enum wrapping.

src/metakb/cli.py
```diff
@@ -363,9 +662,9 @@ def _retrieve_s3_cdms() -> str:
         msg = "Unable to locate files matching expected resource pattern in VICC s3 bucket"
         raise FileNotFoundError(msg)
 
-    echo_info(f"Retrieved CDM files dated {newest_version}")
+    _echo_info(f"Retrieved CDM files dated {newest_version}")
     return newest_version
 
 
 if __name__ == "__main__":
-    update_metakb_db(_anyio_backend="asyncio")
```
jsstevenson, line 371: not totally sure why the asyncio thing was needed? maybe an old click version? I've been able to run the async `transform` command without problems.

README.md
```diff
@@ -114,16 +114,15 @@ MetaKB relies on environment variables to set in order to work.
...
-```sh
+```shell
 pipenv shell
-python3 -m metakb.cli --db_url=bolt://localhost:7687 --db_username=neo4j --db_password=<neo4j-password-here> --load_normalizers_db
+metakb metakb load-normalizers
+metakb update --refresh_source_caches
 ```
 
-For more information on the different CLI arguments, see the [CLI README](docs/cli/README.md).
+The `--help` flag can be provided to any CLI command to bring up additional documentation.
```
jsstevenson, line 125: There's also a way to embed Click docs into Sphinx, so I'll do that in the `docs` branch.

src/metakb/cli.py
```diff
@@ -14,178 +14,414 @@
...
+    :param normalizer: tuple (possibly empty) of normalizer names to check
+    """  # noqa: D301
+    if not check_normalizer_health(db_url, normalizer):
+        _logger.warning("Normalizer check failed.")
+        click.get_current_context().exit(1)
+    _logger.info("Normalizer check passed.")
+
+
+@cli.command()
+@click.option("--normalizer_db_url", "-n", help=_normalizer_db_url_description)
```
jsstevenson, line 120: Since Dynamo takes a single URL but Postgres requires different libpq-style URLs that include normalizer-specific database names, I don't think it's really feasible to support postgres connections this way yet. Thought about doing something that could take a base postgres URL and add database names on top but it's probably not worth the work right now.

src/metakb/cli.py
```diff
@@ -14,178 +14,414 @@
...
 @click.option(
-    "--db_username",
-    default="neo4j",
+    "--refresh_source_caches",
+    "-r",
+    is_flag=True,
+    default=False,
     help=(
-        "Username to provide to application Neo4j database. Can also be provided via environment variable METAKB_DB_USERNAME, which takes priority."
+        "True if source caches (e.g. CivicPy) should be updated prior to data regeneration. Note this will take several minutes. False if local cache should be used"
```
korikuzma, line 194:
```suggestion
        "True if source caches (e.g. CIViCPy) should be updated prior to data regeneration. Note this will take several minutes. False if local cache should be used"
```

src/metakb/cli.py
```diff
@@ -194,136 +430,225 @@ async def update_metakb_db(
...
+@click.option("--db_url", "-u", default="", help=_neo4j_db_url_description)
+@click.option("--db_credentails", "-c", help=_neo4j_creds_description)
+@click.option("--normalizer_db_url", "-n", help=_normalizer_db_url_description)
+@click.option(
+    "--refresh_source_caches",
+    "-r",
+    is_flag=True,
+    default=False,
+    help=(
+        "`True` if source caches (e.g. CivicPy) should be updated prior to data regeneration. Note this will take several minutes. `False` if local cache should be used"
```
korikuzma, line 450:
```suggestion
        "`True` if source caches (e.g. CIViCPy) should be updated prior to data regeneration. Note this will take several minutes. `False` if local cache should be used"
```

korikuzma (changes requested):
nice 🧼

src/metakb/normalizers.py
```diff
@@ -222,3 +232,86 @@ def get_regulatory_approval_extension(
                 )
 
         return regulatory_approval_extension
+
+
+class NormalizerName(StrEnum):
```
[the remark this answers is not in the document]
jsstevenson: Ugh I thought this was a 3.10 thing

src/metakb/normalizers.py
```diff
@@ -222,3 +232,86 @@ def get_regulatory_approval_extension(
...
+                success = False
+        except Exception as e:
+            _logger.error(
+                "Encountered exception while checking %s normalizer: %s", name.value, e
+            )
+            success = False
+    return success
+
+
+def update_normalizer(normalizer: NormalizerName, db_url: str | None) -> None:
```
korikuzma, line 302: Would we want to move the AWS checks here? I don't think MetaKB should be used to update any of the AWS normalizer instances

src/metakb/cli.py
korikuzma, line 1: There are some places where singular name is used for arguments (source, normalizer). It might make sense to make these plural since they can take > 1 value

korikuzma (commented):
Last small thoughts

src/metakb/cli.py
jsstevenson: > There are some places where singular name is used for arguments (source, normalizer). It might make sense to make these plural since they can take > 1 value

@korikuzma yeah agreed that this becomes awkward in code. My sense, FWIW, is that it's more common on the help message itself to make the argument name singular and then represent plurality with ellipses, e.g. `[FILE ...]` or `[FILE] ...`. (See eg [here](http://docopt.org/) or [here](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) or `man ls`). We can use the `metavar` param to specify how the arg is named in help messages while using a more appropriate variable name in the CLI code itself.

src/metakb/cli.py
```diff
@@ -140,31 +154,25 @@ def update_normalizers(
     \f
     :param db_url: URL endpoint of normalizers DynamoDB database. If not given, the
```
korikuzma, line 155:
```suggestion
    :param normalizer_db_url: URL endpoint of normalizers DynamoDB database. If not given, the
```

src/metakb/normalizers.py
```diff
@@ -299,19 +303,36 @@ def check_normalizers(
...
+
+# map normalizer to env var used to designate production DB setting
+normalizer_aws_env_vars = {
+    NormalizerName.DISEASE: DISEASE_AWS_ENV_VAR_NAME,
+    NormalizerName.THERAPY: THERAPY_AWS_ENV_VAR_NAME,
+    NormalizerName.GENE: GENE_AWS_ENV_VAR_NAME,
+}
+
+# map normalizer to update function
+_normalizer_method_dispatch = {
```
korikuzma, line 318: Constants should be in all uppercase

```suggestion
NORMALIZER_AWS_ENV_VARS = {
    NormalizerName.DISEASE: DISEASE_AWS_ENV_VAR_NAME,
    NormalizerName.THERAPY: THERAPY_AWS_ENV_VAR_NAME,
    NormalizerName.GENE: GENE_AWS_ENV_VAR_NAME,
}

# map normalizer to update function
_NORMALIZER_METHOD_DISPATCH = {
```

korikuzma (changes requested):
Small requests

korikuzma (approved):

PR #346, opened 2024-06-04
feat: perform better logging

### LINKED ISSUE #48: Establish logging protocol and consistency
We currently are using Python's built-in [`logging` module](https://docs.python.org/3.8/library/logging.html) in the harvesters to provide information about runtime events. The first question to resolve is whether this is our preferred library (or at least a tentatively acceptable one). Assuming it is, `logging` provides 5 levels of information, and the logging module can be configured to record events up to a certain level (e.g. we may write code to log something at level DEBUG during development, but the production application can be configured to only record events up to WARNING or INFO). [From the official docs](https://docs.python.org/3/howto/logging.html#when-to-use-logging): 

> DEBUG: Detailed information, typically of interest only when diagnosing problems.
> INFO: Confirmation that things are working as expected.
> WARNING: An indication that something unexpected happened, or indicative of some problem in the near future (e.g. ‘disk space low’). The software is still working as expected.
> ERROR: Due to a more serious problem, the software has not been able to perform some function.
> CRITICAL: A serious error, indicating that the program itself may be unable to continue running.

In the normalizers, we've been using INFO to signify phases of computation (like the beginning of an individual source load method) and to mark the beginning/end of long computations (like a download method), and we've used ERROR to note when an expected value is missing (like a reference lookup fails unexpectedly). If an Exception is unhandled and thrown, that will also get caught by logging, but CRITICAL can be used to provide additional context where helpful.

Broadly, I'm not attached to any specific logging method or pattern, but I think we'll save ourselves some future headaches if we can lock some practices down now.

jsstevenson:
arguably close #48

### DIFF
--- a/src/metakb/main.py
+++ b/src/metakb/main.py
@@ -1,15 +1,31 @@
 """Main application for FastAPI."""
+from collections.abc import AsyncGenerator
+from contextlib import asynccontextmanager
 
 from fastapi import FastAPI, Query
 from fastapi.openapi.utils import get_openapi
 
+from metakb.log_handle import configure_logs
 from metakb.query import QueryHandler
 from metakb.version import __version__
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI) -> AsyncGenerator:
+    """Configure FastAPI instance lifespan.
+
+    :param app: FastAPI app instance
+    :return: async context handler
+    """
+    configure_logs()
+    yield
+
 
 app = FastAPI(
     docs_url="/api/v2",
     openapi_url="/api/v2/openapi.json",
     swagger_ui_parameters={"tryItOutEnabled": True},
+    lifespan=lifespan,
 )
 query = QueryHandler()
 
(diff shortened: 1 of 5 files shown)


src/metakb/main.py
```diff
@@ -1,15 +1,31 @@
...
+
+@asynccontextmanager
+async def lifespan(app: FastAPI) -> AsyncGenerator:
+    """Configure FastAPI instance lifespan.
+
+    :param app: FastAPI app instance
+    :return: async context handler
+    """
+    configure_logs()
+    yield
```
jsstevenson, line 21: not convinced that this should live here if we can avoid it, but since this is our entry point for production service, I think we have to force configuration of logs here (ideally we could set it in the lambda runtime client definition

src/metakb/transform/civic.py
```diff
@@ -305,6 +305,7 @@ def _add_variant_therapeutic_response_studies(
                 r["variant_origin"], civic_gene
             )
 
+            _logger.debug("line 308")
```
korikuzma, line 308: is this needed? along with the other `_logger.debug("line #")`?

jsstevenson: Oh thanks. This was what spurred me to try to do the log fixes in the first place.

korikuzma (approved):

PR #352, opened 2024-06-10
fix: use correct Disease model

jsstevenson:
also upgrade validation error logging -- if an entire study result is getting ignored because of a mismatch between the DB data structure and the client schema, that seems like a Big Deal

### DIFF
--- a/src/metakb/query.py
+++ b/src/metakb/query.py
@@ -4,9 +4,9 @@ import logging
 from copy import copy
 from enum import Enum
 
-from disease.query import Disease
 from ga4gh.core._internal.models import (
     Coding,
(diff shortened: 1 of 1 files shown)


src/metakb/query.py
```diff
@@ -4,9 +4,9 @@
 from copy import copy
 from enum import Enum
 
-from disease.query import Disease
```
korikuzma, line 7: oh wow

korikuzma (approved):

src/metakb/query.py
jsstevenson: autoimport fail, the absolute worst

Issue #311, opened 2024-03-28
Streamline graph DB queries


jsstevenson:
It's been true for a while now that some of our core API endpoints require multiple ordered requests to the graph DB to succeed. For example, searching for a study first retrieves that study and its relationships, but can then spur further calls for certain kinds of variations and therapies. This is totally reasonable for now but as requirements and the data models begin to settle, we should be aiming to condense those protocols into a single DB call eventually.

jsstevenson:
It might also be worth looking at whether some queries should be managed transactions vs the new `driver.execute_query` API https://neo4j.com/docs/python-manual/current/query-simple/

jsstevenson:
I think we made the decision to table this one for a while until some of the data models are more solidified -- the changes that I have in mind would be quite brittle to schema updates.

korikuzma:
@jsstevenson I plan on updating the normalizers/ga4gh.vrs and then pulling in cat-vrs/va-spec python model changes.

jsstevenson:
I made some work on this type of problem in https://github.com/cancervariants/metakb/compare/main...new_search but that was... a long time ago and may no longer be needed, or relevant, or both. Will take a look again.

PR #361, opened 2024-06-13
feat: add batch search by variations

### LINKED ISSUE #290: Create a batch get for the metakb API
Hi there! Wanted to get a conversation started for implementing a batch getter for the metakb API.

In contexts like processing a large number of variant IDs from a VCF, it could be helpful to send multiple IDs at once. For reference, on my 2017 MacBook Pro (2.3 GHz Dual-Core Intel Core i5 cores, 8GB memory) with multiprocessing, I max out at ~25 calls per second. Using this, working with a 16MB VCF (one patient, one chromosome) with 330k variants would take ~3.5 hours. The compute doesn't seem to be a problem, as I'm far from maxing out my CPU or memory. Considering this is a relatively small VCF, scaling this to the multi-patient, larger VCFs, and all chromosomes could end up being lengthy regardless of the machine.

Just wanted to provide some info for one use case, but happy to provide more info or discuss this more in detail.

jsstevenson:
close #290 

Current use case is variations only (e.g. in the context of a VCF), but the objects/API are structured in such a way to (relatively) smoothly accommodate other kinds of terms*.

Very naively implemented. Tune-ups in #311 should include fetching everything in a single query rather than iteratively getting every study once study IDs are acquired. I also wonder if there's anything else we can do cache-wise to optimize for those cases where a user is performing repeated lookups on things that are probably pretty closely located in the genome, such as VCFs (seqrepo already uses a big LRU cache for sequence lookups so that might be all that's necessary)

The ability to submit multiple terms for the same kind of entity raises questions about transparent management of redundancy, failed lookups, etc. I added an extra field to supply the normalized ID, if available, for each term, so that the client can tell if terms are normalizing to the same ID or if they fail to normalize at all:

```JSON
  "query": {
    "variations": [
      {
        "term": "EGFR L858R",
        "normalized_id": "ga4gh:VA.S41CcMJT2bcd8R4-qXZWH1PoHWNtG2PZ"
      },
      {
        "term": "matthew cannon 2"
      },
      {
        "term": "ga4gh:VA.S41CcMJT2bcd8R4-qXZWH1PoHWNtG2PZ",
        "normalized_id": "ga4gh:VA.S41CcMJT2bcd8R4-qXZWH1PoHWNtG2PZ"
      }
    ]
  }
```

This also makes it possible to understand which studies correspond to each search term.

* Caveat about including other kinds of entities: I can understand why we'd want it, but it potentially makes search very weird. Say you search for drug A, drug B, variation A, and variation B. Do you just include any study that includes any of those terms? (I don't know what the use case for that is, but I think it's the most intuitive result). Or do you just include studies that include variation A + drug A or drug B, or variation B + drug A or drug B? (actually a pretty reasonable thing to search for, but a bit counterintuitive to frame). 
* A note about response structure. As a first pass, I made it consistent with `get_search_studies` -- `response.studies` is just a list of studies. If you wanted to figure out which ones came from a given search term, you could take the normalized ID from `response.queries` and then filter through `response.studies` based on the value in `study.variant.definingContext.id` (for ProteinSequenceConsequences, or something else for CategoricalVariations). Alternatively, you could group `response.studies` into a dict where the key is the normalized ID and the value is the list of studies for that ID. However, that becomes VERY messy if you *do* want to support searching by multiple entity types (what would the key be, a concatenated string or something?).

### DIFF
--- a/src/metakb/main.py
+++ b/src/metakb/main.py
@@ -102,7 +102,6 @@ async def get_studies(
 
 _batch_search_studies_descr = {
     "summary": "Get nested studies for all provided variations.",
-    "response_description": "todo",
     "description": "Return nested studies associated with any of the provided variations.",
     "arg_variations": "Variations (subject) to search. Can be free text or VRS variation ID.",
 }
--- a/src/metakb/main.py
+++ b/src/metakb/main.py
@@ -117,7 +117,7 @@ async def batch_get_studies(
         None, description=_batch_search_studies_descr["arg_variations"]
     ),
 ) -> dict:
-    """Fetch all studies associated with the provided variations.
+    """Fetch all studies associated with `any` of the provided variations.
 
     :param variations: variations to match against
     :return: batch response object
--- a/src/metakb/main.py
+++ b/src/metakb/main.py
@@ -98,3 +98,30 @@ async def get_studies(
     """
     resp = await query.search_studies(variation, disease, therapy, gene, study_id)
     return resp.model_dump(exclude_none=True)
+
+
+_batch_search_studies_descr = {
+    "summary": "Get nested studies for all provided variations.",
+    "response_description": "todo",
+    "description": "Return nested studies associated with any of the provided variations.",
+    "arg_variations": "Variations (subject) to search. Can be free text or VRS variation ID.",
+}
+
+
+@app.get(
+    "/api/v2/batch_search/studies",
+    summary=_batch_search_studies_descr["summary"],
+    description=_batch_search_studies_descr["description"],
+)
+async def batch_get_studies(
+    variations: list[str] | None = Query(  # noqa: B008
+        None, description=_batch_search_studies_descr["arg_variations"]
+    ),
+) -> dict:
+    """Fetch all studies associated with the provided variations.
+
+    :param variations: variations to match against
+    :return: batch response object
+    """
+    response = await query.batch_search_studies(variations)
+    return response.model_dump(exclude_none=True)
--- a/src/metakb/query.py
+++ b/src/metakb/query.py
@@ -23,6 +23,7 @@ from metakb.schemas.annotation import Document, Method
 from metakb.schemas.api import (
     BatchSearchStudiesQuery,
     BatchSearchStudiesService,
+    NormalizedQuery,
     SearchStudiesService,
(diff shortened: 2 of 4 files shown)


src/metakb/query.py
```diff
@@ -708,3 +714,66 @@ def _get_therapeutic_agent(in_ta_params: dict) -> TherapeuticAgent:
...
+
+    async def batch_search_studies(
+        self,
+        variations: list[str] | None = None,
+    ) -> BatchSearchStudiesService:
+        """Fetch all studies associated with any of the provided variation description
+        strings.
+
+        Because this method could be expanded to include other kinds of search terms,
+        ``variations`` is optionally nullable.
```
korikuzma, line 726: Same comment as above. Can we have it default to empty list?

src/metakb/query.py
```diff
@@ -708,3 +714,66 @@ def _get_therapeutic_agent(in_ta_params: dict) -> TherapeuticAgent:
...
+        """
+        response = BatchSearchStudiesService(
+            query=BatchSearchStudiesQuery(variations=[]),
+            service_meta_=ServiceMeta(),
+            warnings=[],
+        )
+        if not variations:
+            return response
+
+        for query_variation in variations:
```
korikuzma, line 754: Since we're calling the variation-normalizer, we probably should only query unique items

```suggestion
        for query_variation in set(variations):
```

korikuzma (changes requested):

src/metakb/main.py
```diff
@@ -98,3 +98,29 @@ async def get_studies(
...
+
+
+@app.get(
+    "/api/v2/batch_search/studies",
+    summary=_batch_search_studies_descr["summary"],
+    description=_batch_search_studies_descr["description"],
+)
+async def batch_get_studies(
+    variations: list[str] | None = Query(  # noqa: B008
+        None, description=_batch_search_studies_descr["arg_variations"]
```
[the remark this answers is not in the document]
jsstevenson: Does the FastAPI Query method shield from the usual mutable defaults problem?

src/metakb/query.py
jsstevenson: We definitely can't use mutable defaults here, right?

PR #343, opened 2024-05-14
docs: initial docs work

jsstevenson:
There's a lot of TODOs to... do, still, but this pins some rudimentary progress.

* Add docs skeleton, boilerplate
* First stab at setup instructions (they are a doozy)

Major remaining items in future PRs:
* some kind of big sources overview page
* description of harvest/transform procedure, references to VA/VRS
* more description of normalization

PR #365, opened 2024-06-17
fix: querying studies based on therapeutic

### LINKED ISSUE #360: Work out intersection vs union issues in `_get_related_studies` query
- [ ] Is the intention to get all studies that incorporate _any_ of the queried terms, or _all_ of them? I think the latter, but make method name and docstring clearer.
- [ ] Fix behavior when searching for a therapy and other kinds of terms: currently, the `s1 AS s` clause seems to shadow the previously-defined studies earlier in the query, so the disease and variation clauses are ignored. See, for example, the differences between these two queries:

```cql
MATCH (s) -[:HAS_VARIANT] -> (cv:CategoricalVariation)
MATCH (cv) -[:HAS_DEFINING_CONTEXT|HAS_MEMBERS] -> (v:Variation {id:"ga4gh:VA._8jTS8nAvWwPZGOadQuD1o-tbbTQ5g3H"})
MATCH (s) -[:HAS_TUMOR_TYPE] -> (c:Condition {disease_normalizer_id:"ncit:C2926"})
RETURN s
```

and 

```cql
MATCH (s) -[:HAS_VARIANT] -> (cv:CategoricalVariation)
MATCH (cv) -[:HAS_DEFINING_CONTEXT|HAS_MEMBERS] -> (v:Variation {id:"ga4gh:VA._8jTS8nAvWwPZGOadQuD1o-tbbTQ5g3H"})
MATCH (s) -[:HAS_TUMOR_TYPE] -> (c:Condition {disease_normalizer_id:"ncit:C2926"})
MATCH (s1:Study) -[:HAS_THERAPEUTIC] ->(tp:TherapeuticAgent {therapy_normalizer_id:"ncit:C2201"})
RETURN s1 as s
UNION
MATCH (s2:Study) -[:HAS_THERAPEUTIC]-> () -[:HAS_SUBSTITUTES|HAS_COMPONENTS] ->(ta:TherapeuticAgent {therapy_normalizer_id:"ncit:C2201"})
RETURN s2 as s
```

Assuming we want the intersection (incorporate all queried terms) and not union, then the results of the first query should be a superset of the results of the second query, but instead, they return entirely different studies from each other.

korikuzma:
* close #360
* fix cypher query when passed a therapeutic concept


tests/unit/test_search_studies.py
```diff
@@ -1,6 +1,6 @@
-"""Test the MetaKB search_studies method"""
-
+"""Test the MetaKB search_studies_intersect method"""
```
jsstevenson, line 1: this change can be reverted

src/metakb/main.py
```diff
@@ -78,7 +78,7 @@ def custom_openapi() -> dict:
     response_description=search_study_response_descr,
     description=search_studies_descr,
 )
-async def get_studies(
+async def get_studies_intersect(
```
jsstevenson, line 81: now feeling like maybe elongating the method name is unnecessary with actual search behavior ironed out and more explicit docstrings. idk. just seems like a long name.

jsstevenson (approved):

jsstevenson (approved):

korikuzma:
@jsstevenson Am I good to merge or should I wait until #343 is merged?

jsstevenson:
@korikuzma  👍  go ahead

PR #370, opened 2024-06-25
refactor!: remove `extract_harvester` from `transform` method

### LINKED ISSUE #341: Remove `extract_harvester` call from `transform`
We should follow the approach in #338 and remove `extract_harvester` call from `transform`. We should just accept the harvested_data as parameters instead

korikuzma:
close #341

* create new method `extract_harvested_data` to separate these out
* `transform` now accepts harvested data as a parameter

Notes:
* I was considering moving extract harvester to the Harvester class or module. Decided not to for now and may address it in a separate issue


korikuzma:
I'm going to change this to draft until #343 gets merged

jsstevenson (approved):

PR #374, opened 2024-06-28
feat!: replace instance var lists with pydantic model

korikuzma:
I think this is still a wip but @jsstevenson wanted to make you aware


src/metakb/transform/base.py
```diff
@@ -225,14 +249,7 @@ def __init__(
...
 
-        self.studies = []
-        self.categorical_variations = []
-        self.variations = []
-        self.genes = []
-        self.therapeutic_procedures = []
-        self.conditions = []
-        self.methods = []
-        self.documents = []
+        self.processed_data = TransformedData()
```
korikuzma, line 252: Naming is hard. Do we like this? Something else?

korikuzma:
I think refactoring cache strategy should be a separate issue.

jsstevenson (approved):
👍 great. I think it makes a lot of sense to avoid using dicts as an interface layer when we could be using namedtuples (gives some field definitions, enables autocomplete), dataclasses (namedtuples but also enables type annotation) or pydantic models (dataclasses but with runtime validation).

PR #377, opened 2024-07-02
feat!: refactor graph DB code, streamline unneeded query transaction usage

### LINKED ISSUE #373: Refactor graph/DB methods
Started looking at this in #356 but the later versions of the Neo4j Python SDK recommend that you use `Graph.execute_query` for simple queries (for our purposes, reads). Additionally, we're using a transaction-based API for all of our writes, but we're not actually packaging them as collected transactions. Finally, it might be good to separate out concerns related to graph connection instantiation/management and tailored write methods upon the graph.

jsstevenson:
The main performance change here is to do away with querying code that employs transactions where it doesn't need to -- for reads, if subquery 1 (e.g. checking that a study ID is valid) runs successfully, but then something wrong happens in query 2 (e.g. fetching the nested data for that study), we don't need to do anything to the result of query 1 (it already happened). Everything in the `query` module is a read, so we can just use the now-standard `driver.execute_query` method for reads instead of creating a `session` context manager every time. This overhead is probably quite minimal but I think it also simplifies things.

Similarly, when preparing a read-only connection, we probably don't need to check for the existence of constraints, so an option is added to suppress that check.

The `Graph` class is also refactored here to separate responsibilities better between code that acquires/handles DB connections (which remains in the `database` module) and code that transforms knowledgebase data into write queries to store in the DB (moved to the `load_data` (???) module). 

Finally, we weren't really making meaningful use of anything stateful in the Graph code, so it's refactored down to functions, rather than a class. The `QueryHandler` wasn't even using the graph (just the driver), so rather than a class that creates and holds onto a driver, the `get_driver()` function handles credentials processing and provides a Neo4j `Driver` instance to downstream users (the CLI, the QueryHandler).

close #373  (there's more work to do there but fine to cross it out for now especially with possible changes to the data model still incoming)

### DIFF
--- a/docs/source/reference/api/metakb.database.rst
+++ b/docs/source/reference/api/metakb.database.rst
@@ -1,0 +1,8 @@
+﻿metakb.database
+===============
+
+.. automodule:: metakb.database
+   :members:
+   :undoc-members:
+   :special-members: __init__
+   :exclude-members: model_fields, model_config, model_computed_fields
--- a/docs/source/reference/api/metakb.load_data.rst
+++ b/docs/source/reference/api/metakb.load_data.rst
@@ -1,0 +1,8 @@
+﻿metakb.load_data
+================
+
+.. automodule:: metakb.load_data
+   :members:
+   :undoc-members:
+   :special-members: __init__
+   :exclude-members: model_fields, model_config, model_computed_fields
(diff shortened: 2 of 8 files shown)


korikuzma (approved):

korikuzma (commented):
Oh wait, it only showed me doc changes in notifs

korikuzma (approved):
🚀

PR #383, opened 2024-07-15
cicd: update actions versions, stalebot, + templates from template repo

korikuzma:


.github/workflows/stale.yaml
```diff
@@ -1,24 +1,27 @@
...
-    runs-on: ubuntu-latest
-    permissions:
-      issues: write
-      pull-requests: write
-    steps:
-      - uses: actions/stale@v9
-        with:
-          remove-stale-when-updated: true
+  stale-high-priority:
+    uses: genomicmedlab/software-templates/.github/workflows/reusable-stale.yaml
```
jsstevenson, line 9: add @main rev

jsstevenson (changes requested):
tag rev

.github/workflows/stale.yaml
korikuzma: ugh copied/pasted from wrong repo

jsstevenson (approved):

PR #384, opened 2024-07-16
feat: enable pagination

jsstevenson:
* use `start` and `limit` params for pagination
* set default `limit` in `QueryHandler` init
* responses now have order (tbh they might've had it previously in practical terms but this makes it explicit)


tests/unit/search/utils.py
```diff
@@ -0,0 +1,33 @@
...
+        if study.id == expected_study["id"]:
+            actual_study = study
+            break
+
+    if should_find_match:
+        assert actual_study, f"Did not find study ID {expected_study['id']} in studies"
+        resp_studies = [actual_study.model_dump(exclude_none=True)]
+        assertion_checks(resp_studies, [expected_study])
+    else:
+        assert actual_study is None
```
jsstevenson, line 33: could also put these functions into fixtures, but I'm increasingly coming around to the idea that it's cleaner to just put an empty `__init__.py` into a test subdirectory instead of making `conftest.py` a million lines long with a bunch of functions that return other functions

korikuzma:
@jsstevenson not sure if you want me to review, but there are some conflicts to resolve

korikuzma (approved):
Sweet!

PR #386, opened 2024-07-16
build!: update normalizer + ga4gh.vrs depdencies

### LINKED ISSUE #379: Update normalizer/ga4gh dependencies
All of our normalizers and ga4gh.vrs has made new releases with breaking changes. We should update accordingly.

korikuzma:
close #379

### DIFF
--- a/src/metakb/schemas/annotation.py
+++ b/src/metakb/schemas/annotation.py
@@ -4,7 +4,7 @@ import datetime
 from enum import Enum
 from typing import Literal
 
-from ga4gh.core import core_models
+from ga4gh.core.entity_models import IRI, Coding, _DomainEntity, _Entity
 from pydantic import Field, StrictInt, StrictStr, constr, field_validator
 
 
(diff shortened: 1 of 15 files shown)


src/metakb/schemas/annotation.py
```diff
@@ -4,7 +4,7 @@
 from enum import Enum
 from typing import Literal
 
-from ga4gh.core import core_models
+from ga4gh.core.entity_models import IRI, Coding, _DomainEntity, _Entity
```
jsstevenson, line 7: not actionable here, but if we're importing "private" classes like `_DomainEntity` here, maybe they shouldn't be denoted as "private"

jsstevenson (approved):

jsstevenson:
also, fwiw, I am not sure you need to mark this as a breaking change (doesn't really matter though)

korikuzma:
> also, fwiw, I am not sure you need to mark this as a breaking change (doesn't really matter though)

Oh, I thought since the output changed from `alias` to `alternativeLabels` as well as removing `type` from `Extension` field that it would be considered breaking

src/metakb/schemas/annotation.py
korikuzma: Ya I was thinking the same. I thought about inheriting from ABC and removing underscore but haven't made an issue

Issue #409, opened 2024-12-12
MOA disease unique contraint failed


korikuzma:
In #408, I am working on adding support for MOA prognostic data. This resulted in being able to harvest/transform more MOA assertions. During this transformation, I encountered a bug with MOA disease normalization resolution.

```json
{
        "name": "Myelodysplastic Syndromes",
        "oncotree_code": "MDS",
        "oncotree_term": "Myelodysplastic Syndromes"
      }
```

and 

```json
{
        "name": "Myelodysplasia",
        "oncotree_code": "MDS",
        "oncotree_term": "Myelodysplasia"
      }
```

resolve to the same NCIt ID, however we create two separate Disease objects in `MoaTransformer` with the same `id="moa.normalize.disease.ncit:C3247"`, which fails to satisfy the db uniqueness constraint.

Since MOA does not have disease "records" we will have to change how we're processing diseases. One thing I'm not sure about is which value to use for the `label` vs `alternativeLabels`. Typically, we only store normalizer information in the `vicc_normalizer_data` field.

korikuzma:
Ya, I was going back on forth on whether we should handle this or see if MOA could. I drafted a message to Brendan asking if they did any normalization on their side but didn't send and just ended up creating issue here and not sending. I can reach out to him to ask.

korikuzma:
> we just use oncotree for the normalization! the “name” is as written in the source text. we used an underscore in the upcoming release to differentiate “raw” values per a suggestion from James last year

PR #411, opened 2024-12-13
feat!: add support for moa prognostic assertions

### LINKED ISSUE #408: Add support for MOA prognostic data
Currently only MOA therapeutic response assertion data is captured. Now that we have VA-Spec model for prognostic data, we should add the ability to 1) harvest prognostic data 2) transform harvested data to prognostic VA-Spec model 3) load transformed data into db

This is needed for VarCat + ClinVar Submission.

korikuzma:
close #408 

Note: This work revealed a bug with uniqueness constraints (#409). This will be addressed in a separate PR. This PR focuses on changes to the harvester + transformer: `python3 -m pytest tests/unit/harvesters tests/unit/transformers`

* Harvester output changed
  * Therapy fields are now nested inside `therapy` key
  * Remove `clinical_significance` and retain original values from MOA
  * `source_ids` -> `source_id` since we only store one ID
* `MoaTransformer` now supports MOA prognostic assertions

### DIFF
--- a/src/metakb/harvesters/moa.py
+++ b/src/metakb/harvesters/moa.py
@@ -151,24 +151,27 @@ class MoaHarvester(Harvester):
         assertion_record = {
             "id": assertion["assertion_id"],
             "context": assertion["context"],
+            "deprecated": assertion["deprecated"],
             "description": assertion["description"],
             "disease": {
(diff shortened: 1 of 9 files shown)


katiestahl (approved):
looks good to me!

Issue #426, opened 2025-01-31
MOA therapy unique contraint failed


korikuzma:
```json
{
        "name": "Selpercatinib",
        "type": "Targeted therapy",
        "strategy": "RET inhibition",
        "resistance": "",
        "sensitivity": 1
}
```

and 

```json
{
        "name": "LOXO-292",
        "type": "Targeted therapy",
        "strategy": "RET inhibition",
        "resistance": 1,
        "sensitivity": ""
}
```

normalize to the same concept `normalize.therapy.rxcui:2370147`. However, we create two separate Therapy objects in `MoaTransformer` with the same `id="moa.normalize.therapy.rxcui:2370147"`, which fails to satisfy the DB uniqueness constraint. 

This is similar to #409

korikuzma:
Other conflicts:
Abiraterone acetate and Abiraterone resolve to rxcui:1100071
Rapamycin and Sirolimus resolve to rxcui:35302

PR #427, opened 2025-01-31
feat!: add support for transforming concepts that fail to normalize

### LINKED ISSUE #417: Transformers should include concepts that fail to normalize
Currently, we only include concepts that succeed in VICC normalization. However, in the CDMs we also want to be able to include concepts that fail to normalize. For these cases, we'll simply add an extension (`{"name": "vicc_normalizer_failure", "value": True}`).

I'm not really sure how we want to handle this in /search... @mcannon068nw may have some guidance. We can create a separate issue for this. For now, we'll skip loading concepts in the DB that have this extension.

korikuzma:
close #417 

* This adds support for transforming concepts (such as gene, variant, disease, and therapy) that fail to normalize. 
  * Since this work is mainly needed for ClinVar Submission work, this is *only* focused on adding support in CDM files generated by the Transformers. The database will NOT load statements that fail to normalize concepts, and therefore the QueryHandler is unable to search on these non-normalizable concepts. This will require more thought on the search algorithm (exact match? Something else?)
  * Concepts that fail to normalize will have the following `Extension`: `{"name": "vicc_normalizer_failure", "value": True}`
    * Since MOA does not have record IDs for some objects, the IDs used in these cases will be `f"moa.{therapy|disease|gene}:{name}"`, where name is the MOA label for the concept where whitespace characters are replaced with underscores. 

Note:
* Some databases tests will fail due to #426 (I plan to tackle this next)


korikuzma:
Need to look this over before requesting review

katiestahl (approved):

PR #433, opened 2025-02-07
feat!: use system identifiers from source for normalization 

### LINKED ISSUE #431: Do not normalize based on highest match
Based on conversation with @ahwagner, if a source provides an identifier we should use this as the query rather than finding the highest match normalized result.

korikuzma:
close #431 

* Rather than returning the highest match for a list of queries, use system identifiers for source
* Input is now a single query, rather than a list


src/metakb/normalizers.py
```diff
@@ -331,6 +218,41 @@ def get_regulatory_approval_extension(
...
+        except Exception as e:
+            _logger.error(
+                "%s Normalizer raised an exception using query %s: %s",
+                concept_name.capitalize(),
+                query,
+                e,
+            )
+        else:
+            if normalizer_resp.match_type:
+                normalized_id = getattr(normalizer_resp, concept_name).primaryCode.root
```
mcannon068nw, line 252: naive question for my own learning, is this just a hanging `else?` doesn't this normally follow an `if` statement, how does this work?

korikuzma: So for try/except blocks, `else` is executed if no exceptions were raised in the `try` block. More info here: https://www.geeksforgeeks.org/try-except-else-and-finally-in-python/

src/metakb/transformers/civic.py
```diff
@@ -809,6 +810,52 @@ def _get_expressions(self, variant: dict) -> list[Expression]:
                 expressions.append(Expression(syntax=syntax, value=hgvs_expr))
         return expressions
 
+    @staticmethod
+    def _update_normalizer_mappings(
+        mappings: list[ConceptMapping],
+        civic_primary_id: str,
+        get_concept_mapping_method: Callable,
+    ) -> None:
+        """Update ``mappings`` when concepts successfully normalizer
```
mcannon068nw, line 819: normalize?

src/metakb/normalizers.py
```diff
@@ -83,188 +83,75 @@ def __init__(self, db_url: str | None = None) -> None:
...
         >>> from metakb.normalizers import ViccNormalizers
         >>> v = ViccNormalizers()
-        >>> gene_terms = [
-        ...     "gibberish",  # won't match
-        ...     "NETS",  # alias
-        ...     "hgnc:1097",  # HGNC identifier for BRAF
-        ...     "MARCH3",  # previous symbol
-        ... ]
-        >>> v.normalize_gene(gene_terms)[0].normalized_id
+        >>> v.normalize_gene("BRAF")[0].normalized_id
```
mcannon068nw, line 116: is the object that gets returned with normalize_gene() == what gets returned with _normalize_concept() (from the ViccNormalizers class definition). Why would you call both [0] and the .normalized_id here? 

Honestly I think my question is just that I am confused a little bit on what the ViccNormalizers() object looks like and how its intended to be used.

mcannon068nw (dismissed):
A few hanging questions for my own understanding, but all in all I think everything here makes sense to me and looks good to go!

I see that this has been up for a few weeks, sorry for the delay on my end!

src/metakb/normalizers.py
korikuzma: Good call out -- normalized_id was removed (for gene, in [this](https://github.com/cancervariants/gene-normalization/pull/381) pr) so this is actually wrong. I'll go ahead and update

korikuzma:
> I see that this has been up for a few weeks, sorry for the delay on my end!

I was chaining a lot of PRs, so it's been in draft for a bit. I just opened it yesterday haha

katiestahl (approved):

mcannon068nw (approved):

PR #485, opened 2025-04-29
fix(ci): fix deployment, remove pipfile + update seqrepo version

### LINKED ISSUE #476: Leave Pipenv behind
We currently employ some limited Pipenv usage in this repo for reasons that, if I'm not mistaken, relate to Elastic Beanstalk deployment. It'd be nice to bail on EB too, but in the meantime, we might want to at least get away from hauling around an extra build tool. My gloss of the EB docs suggests that we could just be providing a `requirements.txt` file.

korikuzma:
close #476 

Followed https://github.com/cancervariants/fusion-curation


README.md
```diff
@@ -109,12 +109,6 @@ First, [follow these instructions for deploying DynamoDB locally on your compute
 java -Djava.library.path=./DynamoDBLocal_lib -jar DynamoDBLocal.jar -sharedDb
 ```
 
-Next, navigate to the `site-packages` directory of your virtual environment. Assuming Pipenv is installed to your user directory, this should be something like:
-
-```sh
-cd ~/.local/share/virtualenvs/metakb-<various characters>/lib/python<python-version>/site-packages/  # replace <various characters> and <python-version>
-```
-
```
korikuzma, line 117: I'm not sure this really helped with anything, so I just removed

requirements.txt
katiestahl, line 1: should we make sure to have a note somewhere saying not to manually modify this since it's for deployment? (only use pip freeze after updating deps in pyproject.toml)?

katiestahl (approved):
looks good to me! left a comment on the requirements.txt but it's nothing major that would make me not approve. nice work :)

requirements.txt
katiestahl: also, might be neat to switch to `pip-compile` for updating this file anyways. It auto generates a comment at the top which would help as well!

korikuzma: > also, might be neat to switch to pip-compile for updating this file anyways. It auto generates a comment at the top which would help as well!

If we do want to go this route, I think it should be done in a separate issue since it introduces a new dep

katiestahl (approved):

PR #490, opened 2025-05-05
fix: remove preinstall check since it prevents adding new deps

katiestahl:
I initially added this script to prevent people from using `pnpm install` in the `client` directory, since it's only supposed to be run from the root of a monorepo I believe. Running it from client can cause you to have a `node_modules` folder in both the `root` and `client` directories, which is not ideal. I don't know if there are any potential issues with that, like conflicting dependencies. I don't think that's possible because pnpm looks at all workspaces defined in the workspace yaml, but I added the check script as a failsafe...

BUT this prevents developers from being able to add any dependencies to the `package.json` in `client`, which was not an intended consequence. I think it's best to remove this for now and figure out something later.


katiestahl:
> > I initially added this script to prevent people from using pnpm install in the client directory, since it's only supposed to be run from the root of a monorepo I believe.
> 
> Should we also add this to the readme?

So I was actually just wrong and you should be able to `pnpm install` from anywhere 😓 oops. Apologies -still learning how monorepos are supposed to work

PR #494, opened 2025-06-30
feat: add python-dotenv

jsstevenson:
Add `python-dotenv` to backend entry points.

This is a simple quality-of-life improvement driven by my unending frustration with the mountain of environment variables upon which metakb sits. Store a `.env` file in `server/` and it'll load env vars from there.

* I have aspirations for a more featureful configuration experience. I'd made an initial attempt [here](https://github.com/GenomicMedLab/software-templates/pull/78) but I'm not totally happy with it. I have an issue [here](https://github.com/GenomicMedLab/software-templates/issues/106) to take up someday where I try and do something better, but I think that would entail implementation in all of the library dependencies first. For now I think this PR is just fine though.

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -1,12 +1,10 @@
+METAKB_EB_PROD=
 METAKB_DB_URL=bolt://localhost:7687
 METAKB_DB_USERNAME=neo4j
 METAKB_DB_PASSWORD=neo4j
 METAKB_DB_SECRET=
-DISEASE_AWS_ENV_VAR_NAME=disease_normalizer
 DISEASE_NORM_DB_URL=http://localhost:8000
-THERAPY_AWS_ENV_VAR_NAME=therapy_normalizer
 THERAPY_NORM_DB_URL=http://localhost:8000
-GENE_AWS_ENV_VAR_NAME=gene_normalizer
 GENE_NORM_DB_URL=http://localhost:8000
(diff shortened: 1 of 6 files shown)


jsstevenson:
priority: medium because this really annoys me

korikuzma (dismissed):
I liked that AnyVar had an [example env file](https://github.com/biocommons/anyvar/blob/main/.env.example) for people to reference. My only suggestion would be to add one here.

server/.env.example
korikuzma, line 1: I don't think the `{GENE/DISEASE/THERAPY}_AWS_ENV_VAR_NAME`s are used anymore (tried searching the repos and not seeing anything). Everything else looks good to me!

korikuzma: Oh I see. These are in the normalizers module in MetaKB. I think if we wanted to use these, it would be `{GENE/DISEASE/THERAPY}_NORM_ENV`

jsstevenson: Actually yeah I'm just gonna pull those out because the normalizers are using the "<NAME> in environ" pattern that would be triggered by an empty .env setting

korikuzma: Probably. I never use this section of the codebase since I use existing DDB tables. 

Taking a quick glance, `update_normalizer` function should be updated since it's using old args (`--update_all` + `--update_merged` vs `--all` + `--normalize`)

PR #502, opened 2025-07-10
feat!: locate data dir outside of library space, enable configuration

### LINKED ISSUE #495: Move data contents out of `server/src/metakb/data` + enable configuration of this location
### Feature description

Generated data, such as harvester results, should live somewhere other than their current location (`server/src/metakb/data`) by default. The location should be configurable both within code (ie method calls) and also via an environment variable. There should be some consistency about how this configuration works.

### Use case

A few possible arguments for moving the default location:
* Centralizing their location somewhere means multiple processes can easily and consistently access them without additional configuration. Maybe I want to run some analysis on the metakb data, or I have some additional service that uses it -- if there's a code-defined default location then I can just import that value
* Less stuff to .gitignore
* Theoretically it can be erroneous to be referring to locations relative to module files in this way, although I think this risk is quite limited in practice

As for configuration, I think that you can generally set alternate locations within code itself, e.g. in `Harvester.save_harvested_data_to_file`. But there's no way to configure this value globally (instead you have to do it as an arg for that method, as an arg for the Transformer constructor, as individual file paths for the `metakb load-cdm` command, 

### Acceptance Criteria

* An importable value, defined somewhere, that says where metakb data is. The default value for this should be somewhere in userspace outside of the metakb source code directory. That default should be configurable via env variable.
* (stretch goal) a bit more consistency in the user experience of configuring data locations. It's a little fragmented right now but that might just be a consequence of how different stuff works and maybe wouldn't be very easy to improve.

### Proposed solution

_No response_

### Alternatives considered

In other libraries, we've used approaches similar to wags-tails (ie put them in `~/.local/share/`) or within wags-tails itself (ie something like `~/.local/share/wags-tails/metakb`)

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

Yes, I can create a PR for this feature.

jsstevenson:
close #495

Designate a root metakb data dir that is, by default, outside of app-space (`~/.local/share/wags-tails/metakb/`). Support global configuration of this value via `METAKB_DATA_DIR` env var. Otherwise, fall back on wags-tails default behavior. All of this is still preempted if you pass a specific path value to a given function for saving CDM or harvest or whatever.

* Strictly speaking, this is API-breaking because it alters some arg defaults and removes the `APP_ROOT` variable, but behavior should be pretty similar plus additional config options

Create the `metakb.config` module. Currently it's only used to configure the data directory setting. Rather than using hard-coded environment settings like in the lab template, takes a more pared-back approach to avoid committing to any structural stuff that might get changed later. I think it would be pretty easy to just move this to `pydantic-settings`, for example. I also didn't move any existing environment configs (eg DB username/password) into this module yet.

Add a small docs page about this.

Because they were written a long time ago, the workflow of how file locations are specified and looked up in the harvester/transformer/upload pipeline is a little funky but not too bad. I had labeled this a 'stretch goal' in the ticket but I did not make any changes here beyond changing default behavior.


server/src/metakb/harvesters/base.py
```diff
@@ -54,18 +55,20 @@ def harvest(self) -> _HarvestedData:
...
-        """Save harvested data to JSON file
+        """Save harvested data to JSON file.
 
         :param harvested_data: harvested data from a source
-        :param harvested_filepath: Path to the JSON file where the harvested data will
-            be stored. If not provided, will use the default path of
-            ``<APP_ROOT>/data/<src_name>/harvester/<src_name>_harvester_YYYYMMDD.json``
+        :param harvested_filepath: Path to the JSON file location where the harvested data
+            will be stored. If not provided, will use the default path of
+            ``<METAKB_DATA_ROOT>/<src_name>/harvester/<src_name>_harvester_YYYYMMDD.json``,
```
korikuzma, line 63: Should `METAKB_DATA_ROOT` be `METAKB_DATA_DIR`?

korikuzma (approved):

server/src/metakb/harvesters/base.py
jsstevenson: Oh yeah, I saw this comment and it made me think a bit more about what I was trying to say with these docstrings... will tweak a bit

korikuzma (approved):

PR #503, opened 2025-07-17
feat: add service info endpoint

### LINKED ISSUE #289: Create service info endpoint
Conform to https://www.ga4gh.org/product/service-info/

Info to provide:

- [ ] VRS schema version
- [ ] UTA version
- [ ] SeqRepo version
- [ ] URL/identifier for CDM
- [ ] more?

Other things

- [ ] Write GitHub action to test endpoint adherence to schema

jsstevenson:
close #289 

* Add `GET /service-info` endpoint
* General FastAPI module cleanup

Some other dependency versions (VA-spec, Cat-VRS) aren't available until we upgrade those libraries. There are some things we could try to expose UTA and SeqRepo versions but I think they'd take some more code. I'd propose to spike those to a separate ticket focused on what other metadata should be exposed and how to do it.

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -51,7 +51,7 @@ tests = [
     "mock",
     "pytest-asyncio",
     "deepdiff",
-    "jsonschema"
+    "jsonschema~=4.24",  # pin to avoid deprecation of ref resolver
 ]
 dev = [
     "pre-commit>=4.2.0",
--- a/server/tests/unit/test_api.py
+++ b/server/tests/unit/test_api.py
@@ -15,6 +15,9 @@ def client():
     return TestClient(app)
 
 
+@pytest.mark.filterwarnings(
+    "ignore:jsonschema\\.RefResolver is deprecated:DeprecationWarning"
+)
 def test_service_info(client: TestClient, test_data_dir: Path):
     response = client.get("/service-info")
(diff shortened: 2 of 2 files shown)


jsstevenson:
Upgrading this to 'medium' because it includes some changes to app initialization that would be good to get in place/build upon

server/src/metakb/schemas/api.py
```diff
@@ -8,11 +9,78 @@
...
+class SpecMetadata(BaseModel):
+    """Define substructure for reporting specification metadata."""
+
+    vrs_version: str = VRS_VERSION
+
+
+class ImplMetadata(BaseModel):
+    """Define substructure for reporting metadata about internal software dependencies."""
+
+    vrs_python_version: str = vrs_python_version
```
korikuzma, line 56: We should also include cat-vrs/cat-vrs-python and va-spec/va-spec-python (which made me realize, I forgot to bump this in the latest va-spec-python release)

korikuzma (changes requested):

server/src/metakb/schemas/api.py
korikuzma: Oh JK. I just saw the PR description.

korikuzma (approved):

PR #518, opened 2025-07-30
feat!: update backend schema to support expanded catvar searches

jsstevenson:
Update graph schema to support catvar-based searches. Adds a node for constraints. Cleans up some upload and data retrieval queries.

### DIFF
--- a/server/src/metakb/load_data.py
+++ b/server/src/metakb/load_data.py
@@ -202,73 +202,6 @@ def _add_therapy(tx: ManagedTransaction, therapy_in: dict) -> None:
     tx.run(query, **therapy)
 
 
-def _add_location(tx: ManagedTransaction, location_in: dict) -> None:
-    """Add location node and its relationships
-
-    :param tx: Transaction object provided to transaction functions
-    :param location_in: Location CDM object
-    """
-    loc = location_in.copy()
-    loc_keys = [
-        f"loc.{key}=${key}"
-        for key in ("id", "digest", "start", "end", "sequence", "type")
-        if loc.get(key) is not None  # start could be 0
-    ]
-    loc["sequenceReference"] = json.dumps(loc["sequenceReference"])
-    loc_keys.append("loc.sequenceReference=$sequenceReference")
-    loc_keys = ", ".join(loc_keys)
-
-    query = f"""
-    MERGE (loc:{loc["type"]}:Location {{ id: '{loc["id"]}' }})
-    ON CREATE SET {loc_keys}
-    """
-    tx.run(query, **loc)
-
-
-def _add_variation(tx: ManagedTransaction, variation_in: dict) -> None:
-    """Add variation node and its relationships
-
-    :param tx: Transaction object provided to transaction functions
-    :param variation_in: Variation CDM object
-    """
-    v = variation_in.copy()
-    v_keys = [
-        f"v.{key}=${key}" for key in ("id", "name", "digest", "type") if v.get(key)
-    ]
-
-    expressions = v.get("expressions", [])
-    for expr in expressions:
-        syntax = expr["syntax"].replace(".", "_")
-        key = f"expression_{syntax}"
-        if key in v:
-            v[key].append(expr["value"])
-        else:
-            v_keys.append(f"v.{key}=${key}")
-            v[key] = [expr["value"]]
-
-    state = v.get("state")
-    if state:
-        v["state"] = json.dumps(state)
-        v_keys.append("v.state=$state")
-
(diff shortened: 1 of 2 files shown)


server/src/metakb/query.py
```diff
@@ -613,46 +620,54 @@ def _get_disease(self, node: dict) -> MappableConcept:
...
+            msg = f"Unrecognized set of sequence expression node labels: {state_node.labels}"
+            raise ValueError(msg)
+
+        location = SequenceLocation(
+            start=location_node["start"],
+            end=location_node["end"],
+            id=location_node["id"],
+            sequenceReference={"refgetAccession": location_node["refget_accession"]},
+            sequence=location_node["sequence"],
+            digest=location_node.get("digest"),
```
korikuzma, line 652: Maybe silly question. Are there cases where we wouldn't have a digest for a location?

korikuzma: > FWIW, I would actually prefer to remove this property from the DB nodes, since it's effectively redundant information over the ID. I kept it in because it was gonna be effort to change the test cases and the tests were kinda driving me crazy.

This makes sense to me for locations. Wanna create an issue for this (if it doesn't already exist somewhere)?

server/src/metakb/load_data.py
```diff
@@ -280,44 +353,18 @@ def _add_categorical_variant(
     :param categorical_variant_in: Categorical variant CDM object
```
katiestahl, line 353:
```suggestion
    :param catvar: Categorical variant CDM object
```

server/src/metakb/load_data.py
```diff
@@ -202,76 +202,149 @@ def _add_therapy(tx: ManagedTransaction, therapy_in: dict) -> None:
...
-    if loc:
-        _add_location(tx, loc)
-        query += f"""
-        MERGE (loc:{loc["type"]}:Location {{ id: '{loc["id"]}' }})
-        MERGE (v) -[:HAS_LOCATION] -> (loc)
-        """
-
-    tx.run(query, **v)
+    # catvars currently support a single constraint
+    constraint = catvar["constraints"][0]
```
katiestahl, line 326: I know in `add_categorical_variant` below, you check to make sure `constraint` exists/is nonempty - are there other potential entry-points for this function in which we should validate that in this function as well?

server/src/metakb/load_data.py
```diff
@@ -202,76 +202,149 @@ def _add_therapy(tx: ManagedTransaction, therapy_in: dict) -> None:
...
-    query = f"""
-    MERGE (loc:{loc["type"]}:Location {{ id: '{loc["id"]}' }})
-    ON CREATE SET {loc_keys}
+    :param allele: allele object from CDM
+    :return: reformatted object to better fit DB upload
     """
-    tx.run(query, **loc)
-
+    allele_to_upload = {
+        "id": allele["id"],
```
katiestahl, line 212: do we do any sort of validation for `allele` prior to this function? Should we make sure these attributes are available or add any validation? I don't feel strongly at all if you tell me "no", I just figured I should ask since we assume these keys will be available

mcannon068nw:
> > This PR doesn't change previous behavior: that searching by variation works by trying to match against a known member or constraint allele. But do we need to also materialize more members or project query variations up to the protein level?
> 
> I'm going to defer to @mcannon068nw on this one (and whether this is needed for prototype)

Projecting query variations up to the protein level is not needed for the prototype; we should just demonstrate that we can retrieve and aggregate on the allele constraint. 

Is this something we can leave ourselves room for now to address post-prototype? (along with additional, more complex restraint types)

server/src/metakb/load_data.py
jsstevenson: These objects are basically serialized-then-deserialized pydantic objects, so we do have some guarantees about what's inside them (with the caveat that an evil demon could've slipped in and messed with them during that transition). What I'd like to be doing for these functions is to have them take realized Pydantic classes as input so those guarantees are stricter -- that's part of the big storage arch epic i'm pitching (linked below, described on call today)

katiestahl (approved):
looks good to me :)

jsstevenson:
> Projecting query variations up to the protein level is not needed for the prototype; we should just demonstrate that we can retrieve and aggregate on the allele constraint.
> 
> Is this something we can leave ourselves room for now to address post-prototype? (along with additional, more complex restraint types)

Up to you. The gist of the issue is like -- if a source doesn't say that <genomic variation> is a member of <protein variation>, then a user can't enter the former to get to the latter. A lot of those variations ARE realized in the source data, but it's inconsistent. I'll make another issue.

PR #541, opened 2025-08-15
feat: audit nodes + remove/reshape unneeded properties

### LINKED ISSUE #533: Audit DB node properties for inessential data
For example, `digests` probably don't need to be retained from inputs, and can be dropped.

jsstevenson:
close #533 

Small alterations to graph properties

* remove `conceptType` from gene/disease/therapy nodes
* remove `digest` from allele/sequencelocation nodes (note: this also affects the eventual API output)
* allele `expressions` are just a jsonblob, not flattened properties on the node itself

### DIFF
--- a/server/src/metakb/load_data.py
+++ b/server/src/metakb/load_data.py
@@ -124,7 +124,7 @@ def _add_gene_or_disease(
         raise TypeError(msg)
 
     obj["conceptType"] = obj_type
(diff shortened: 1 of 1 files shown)


jsstevenson:
...ah, I'd hoped this would be a simple fix, but it's gonna entail a lot of changes. I think that maybe htis just has to be included in the big graph refactor.

PR #555, opened 2025-09-08
feat!: restructure /search results to support easier client grouping

### LINKED ISSUE #546: Add statement grouping logic to `/search_statements`
### Feature description

Since `/search_statements` is a service for the UI and not strictly about providing access to a resource, it makes sense to perform some complicated grouping up front and tailor the response structure to make life easier for the client. In this case, we want to a) break out statements by proposition type, and b) perform some sort of grouping rule so that statements with the same variant/therapy/disease are clumped together.

### Use case

It's for the big category result tables.

### Acceptance Criteria

Might need to fill this in more, but generally:

- [ ] Group statements by proposition type
- [ ] Further group them by associated entity IDs

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

Yes, I can create a PR for this feature.

jsstevenson:
close #546 

* add proposition type and entity grouping to statement results. I don't think this is necessarily a permanent solution but it'll get us going on what we need for the UI, and we can bridge from there.
* Add additional structure to the way search term results are reported

Here's a screenshot that kinda captures what's going on

Screenshot 2025-09-09 at 2 15 28 PM
 
* ~Remove `/batch_search`. I think this should probably just be part of the main `/search_statements` endpoint if it's something that we really need.~ Actually I added it back! I do think it should just be merged into the other endpoint though.
* Some internal refactoring to simplify edges between different layers. Do more API stuff in the API endpoint module, and make the query module methods a bit more generalizable.
* Add start/limit params to the response
* I noticed a few cases where edge case/error behavior was being handled in ways contrary to the docstring and fixed this up 

**Question**: I don't remember if we wanted to be grouping by the source entity ID (eg `civic.mpid:12`) or the normalized ID (`ga4gh:VA.abcdefghijkl`)
**Observation for later**: I think in the future it might make more sense to revert back to the original unstructured array of statements, with an option to _also_ add back arrays of statement IDs grouped by different entities. so you get an array of all statements with gene X, or therapy Y, or direction Z etc. This would let the client do more dynamic filtering/grouping on its own. Could make a separate discussion or ticket in the future. This is roughly what I was thinking

```json
{
  "query": ["query terms here"],
  "start": 0,
  "limit": null,
  "statements": ["big list of statements here, including contained diseases and variants etc"],
  "filter_elements": {
    "proposition_type": {
      "therapeutic_response": ["civic.eid:1", "civic.eid:5"],
      "diagnostic": ["moa:1", "civic.eid:2"]
    }
    "therapy": {
      "imatinib": ["civic.eid:1"],
      "trastuzumab": ["civic.eid:5"]
    },
    "condition": {
      "neuroblastoma": ["civic.eid:1", "civic.eid:2"],
    },
    "variation": {
      "ga4gh:VA.abcdefg": ["civic.eid:2"],
      "ga4gh:VA.zyxwvutsrqp": ["civi.eid:5", "civic.eid:1"]
    }
  }
}
```

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,7 +1,7 @@
 name: checks
 on: [push, pull_request]
 jobs:
-  pip_install:
+  test_backend:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
@@ -24,6 +24,10 @@ jobs:
 
       - name: Install dependencies
         run: uv sync --extra tests --extra dev --extra notebooks
+
+      - name: Run tests
+        working-directory: ./server
+        run: uv run pytest -m ci
 
   lint_backend:
     name: lint
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -5,7 +5,9 @@ authors = [
     {name = "Kori Kuzma"},
     {name = "James Stevenson"},
     {name = "Brian Walsh"},
-    {name = "Jeff Liu"}
+    {name = "Jeff Liu"},
+    {name = "Matthew Cannon"},
+    {name = "Katie Stahl"},
 ]
 readme = "README.md"
 classifiers = [
@@ -96,6 +98,9 @@ where = ["src"]
 [tool.pytest.ini_options]
 addopts = "--cov=src --cov-report term-missing --ignore=codebuild/"
 testpaths = ["tests"]
+markers = [
+    "ci: tests safe to run in CI",
+]
 
 [tool.coverage.run]
 branch = true
--- a/server/src/metakb/load_data.py
+++ b/server/src/metakb/load_data.py
@@ -5,13 +5,150 @@ import logging
 import uuid
 from pathlib import Path
 
-from ga4gh.va_spec.base import MembershipOperator
-from neo4j import Driver, ManagedTransaction
+from ga4gh.core.models import Extension, MappableConcept
+from ga4gh.va_spec.base import (
+    ConditionSet,
+    MembershipOperator,
+    Statement,
+    TherapyGroup,
+    VariantDiagnosticProposition,
+    VariantPrognosticProposition,
+    VariantTherapeuticResponseProposition,
+)
+from neo4j import Driver, ManagedTransaction, Transaction
 
 from metakb.database import get_driver
 from metakb.transformers.base import NormalizerExtensionName
 
 _logger = logging.getLogger(__name__)
+
+
+def _has_normalize_failure(extensions: list[Extension] | None) -> bool:
+    """Check if extensions contain indication of normalization failure
+
+    :param extensions: list of extensions from an entity object
+    :return: True if contains normalize failure
+    """
+    if extensions:
+        for extension in extensions:
+            if extension.name == "vicc_normalizer_failure" and extension.value:
+                return True
+    return False
+
+
+def _is_loadable_condition(
+    condition: MappableConcept | ConditionSet, statement_id: str
+) -> bool:
+    """Check if condition or condition set can be loaded into DB
+
+    :param condition: condition or conditionset to check
+    :param statement_id: ID to use in logging failure message
+    :return: whether condition is loadable
+    :raise ValueError: if unsupported type of condition (eg IRI ref) is provided
+    """
+    if isinstance(condition, MappableConcept):
+        if _has_normalize_failure(condition.extensions):
+            _logger.info(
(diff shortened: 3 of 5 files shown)


server/src/metakb/main.py
```diff
@@ -141,6 +133,46 @@ async def get_statements(
...
+        "diagnostic_statements": {},
+        "prognostic_statements": {},
+    }
+    statement_ids = []
+    for statement in search_results.statements:
+        statement_ids.append(statement.id)
+        proposition_type = statement.proposition.type
+        variant_id = statement.proposition.subjectVariant.id
+        predicate = statement.proposition.predicate
+        if proposition_type == "VariantTherapeuticResponseProposition":
```
korikuzma, line 150: Consider changing to 

```python
isinstance(proposition, VariantTherapeuticResponseProposition)
```

server/src/metakb/main.py
```diff
@@ -141,6 +133,46 @@ async def get_statements(
...
+            key = f"{variant_id}|{statement.proposition.conditionQualifier.root.id}|{statement.proposition.objectTherapeutic.root.id}|{predicate}"
+            grouped_statements["tr_statements"].setdefault(key, []).append(statement)
+        elif proposition_type == "VariantDiagnosticProposition":
+            key = f"{statement.proposition.subjectVariant.id}|{statement.proposition.objectCondition.root.id}|{predicate}"
```
korikuzma, line 154:
```suggestion
            key = f"{variant_id}|{statement.proposition.objectCondition.root.id}|{predicate}"
```

server/src/metakb/main.py
```diff
@@ -141,6 +133,46 @@ async def get_statements(
...
+            grouped_statements["diagnostic_statements"].setdefault(key, []).append(
+                statement
+            )
+        elif proposition_type == "VariantPrognosticProposition":
+            key = f"{statement.proposition.subjectVariant.id}|{statement.proposition.objectCondition.root.id}|{predicate}"
```
korikuzma, line 159:
```suggestion
            key = f"{variant_id}|{statement.proposition.objectCondition.root.id}|{predicate}"
```

server/src/metakb/query.py
```diff
@@ -199,196 +196,129 @@ async def search_statements(
...
+
+        # return early if ANY search terms fail to resolve
+        if any(
+            (
+                normalized_therapy and normalized_therapy.resolved_id is None,
+                normalized_disease and normalized_disease.resolved_id is None,
+                normalized_variation and normalized_variation.resolved_id is None,
+                normalized_gene and normalized_gene.resolved_id is None,
+            )
+        ) or (statement_term and statement_term.resolved_id is None):
```
korikuzma, line 239: Could simplify this

```suggestion
        if any(
            obj and obj.resolved_id is None
            for obj in (
                normalized_therapy,
                normalized_disease,
                normalized_variation,
                normalized_gene,
                statement_term,
            )
        ):
```

server/src/metakb/query.py
```diff
@@ -199,196 +196,129 @@ async def search_statements(
...
-        _, normalized_gene_id = self.vicc_normalizers.normalize_gene(gene)
-        if not normalized_gene_id:
-            warnings.append(f"Gene Normalizer unable to normalize: {gene}")
-        return normalized_gene_id
+        # Check if VRS variation (allele, copy number change, copy number count)
+        # TODO should this happen before the normalizer call?
+        if not normalized_variation and variation.startswith(
+            ("ga4gh:VA.", "ga4gh:CX.", "ga4gh:CN.")
+        ):
+            normalized_variation = variation
```
korikuzma, line 316: Yes, I do think if they're passing a VRS ID (do we want or need to support this?), we should do this first and not attempt to normalize.

server/src/metakb/query.py
```diff
@@ -750,14 +680,14 @@ def _get_gene_context_qualifier(self, statement_id: str) -> MappableConcept | No
...
         results = self.driver.execute_query(query, statement_id=statement_id)
         if not results.records:
-            logger.error(
+            _logger.error(
                 "Unable to complete gene context qualifier lookup for statement_id %s",
                 statement_id,
             )
             return None
         if len(results.records) > 1:
             # TODO should this be an error? can statements have multiple gene contexts?
```
korikuzma, line 689: rip me

korikuzma (changes requested):

PR #576, opened 2025-09-29
feat: add deployment scripts and mount static files in backend

### LINKED ISSUE #475: Set up deployment for UI site
I imagine there may be potential new costs associated with deploying a new frontend site somewhere, and since the work in #391 is to restructure into a monorepo with placeholder stuff for the frontend, I didn't want to risk implementing this and having some extra costs for a "Hello World" site.

Creating this ticket so we can track setting up automated deployment for the frontend portion of the site

- how many environments? Should we have a staging and production environment that auto-deploy upon merges to staging or main? Or use release tags?

jsstevenson:
close #475 

* Add elasticbeanstalk deployment scripts for frontend build
* Update server env to include a default 'local' setting
* Refactor routes and dependencies into a `/restapi` subdirectory. As we build out entity-specific endpoints I envision us adding a new `server/src/metakb/restapi/resources.py` module or something to that effect
* Add route/etc to provide static file service
* Use CORS configs copy/pasted from fusion-curation. I hope this means it'll work, but who knows how correct it is. I'd like to make a separate ticket to double check it from a security perspective afterwards.
* Add Makefile w/ targets for generating zips for manual deploy in elasticbeanstalk (hopefully not something we ever need to do again, but it was helpful while writing this and I figured I shouldn't throw it away) and for regenerating `requirements.txt`

The ElasticBeanstalk configs should be updated to provide a `METAKB_ENV` variable with either `"dev"`, `"staging"`, or `"prod"` as the value (lower-case).  I think we should also set `METAKB_DEBUG=true` in staging and dev environments.

For the moment, I've been manually uploading source contents to the dev env, you can see at

* http://metakb-dev-eb.us-east-2.elasticbeanstalk.com/ for client (mostly nonfunctional right now)
* http://metakb-dev-eb.us-east-2.elasticbeanstalk.com/api/v2 for swagger docs (will crash your computer if you try to do anythign)
* http://metakb-dev-eb.us-east-2.elasticbeanstalk.com/api/v2/search/statements?disease=nsclc for an example of an API request


.ebextensions/options.config
```diff
@@ -0,0 +1,3 @@
+option_settings:
+  aws:elasticbeanstalk:application:environment:
+    NODE_ENV: production
```
jsstevenson, line 3: I don't know what this actually does or why it's here, but it was in fusion-curation, soooo

korikuzma (approved):

korikuzma:
This was in a slack message, but once this is merged I will terminate the dev instance

PR #580, opened 2025-10-02
feat: add search results prototype

katiestahl:


korikuzma:
@katiestahl could you link the relevant issues to this PR?

korikuzma:
I added @mcannon068nw because it might be good to see if he has any requests that should be made before merging

client/src/utils/propositions.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  return null
+}
+
+/**
+ * Extracts human-readable condition names from a Condition object.
+ *
+ * @param condition - A string, MappableConcept, ConditionSet, or undefined
+ * @returns Array of condition names (may be multiple for ConditionSet)
```
katiestahl, line 111: @jsstevenson this is what we're currently discussing in the meeting - here's how I decided to handle condition sets - I make every condition a set, whether it's a single condition or a conditionset. Then, these get flattened to create the filters later

jsstevenson: Works for now. I know that coming down the pike we have cases like "Cancer AND (baby onset OR teenage onset)" but we don't necessarily need to have perfect solutions for those cases immediately

client/src/pages/Results/GeneResult.tsx
```diff
@@ -0,0 +1,368 @@
...
+                {description}
+              </Typography>
+            </Box>
+            <Box
+              id="results-table-container"
+              sx={{ backgroundColor: 'white', padding: 5, borderRadius: 2, marginTop: 2 }}
+            >
+              <Tabs
+                onChange={(_, value) => setActiveTab(value)}
+                value={activeTab}
```
jsstevenson, line 275: I think we should (either with a mapping or just manually) make this value be capitalized in line with the context, so that it reads eg "Diagnostic Search Results" instead of "diagnostic Search Results"
Screenshot 2025-10-06 at 4 10 53 PM

client/src/pages/Results/GeneResult.tsx
```diff
@@ -0,0 +1,368 @@
...
+                <Tab
+                  label={`Therapeutic (${filteredByTab.therapeutic.length})`}
+                  value="therapeutic"
+                />
+                <Tab label={`Diagnostic (${filteredByTab.diagnostic.length})`} value="diagnostic" />
+                <Tab label={`Prognostic (${filteredByTab.prognostic.length})`} value="prognostic" />
+              </Tabs>
+              <Typography variant="h6" mb={2} fontWeight="bold">
+                {activeTab} Search Results ({filteredResults?.length})
+              </Typography>
```
jsstevenson, line 287: I think this should be in a div with 16px of padding just like the stuff above and below it

Screenshot 2025-10-06 at 4 12 35 PM

jsstevenson (approved):
A few aesthetic thoughts that could be resolved, spiked, or ignored. Code looks good to me

PR #615, opened 2025-10-22
fix: pin static version value

### LINKED ISSUE #613: Version is unavailable in deployed library
### Describe the bug

Service-info API is unable to provide a `version` value when deployed to beanstalk because it's losing the SCM stuff

### Steps to reproduce

```
curl -X 'GET' \
  'http://metakb-dev-eb.us-east-2.elasticbeanstalk.com/api/v2/service-info' \
  -H 'accept: application/json'
```

### Expected behavior

Should be a real version

### Current behavior

`unknown`

### Acceptance Criteria

- [ ] Make a version value work

### Possible reason(s)

_No response_

### Suggested fix

* Set a version value manually, like we used to
* See if there's a way for codebuild to carry across the right value

### Branch, commit, and/or version

`staging`

### Screenshots

_No response_

### Environment details

asdf;lkjs

### Additional details

_No response_

### Contribution

None

jsstevenson:
close #613 (for now)

* since setuptools-scm gets a version value from a local .git repo, but this isn't included in the deployed product, then there's no way to provide a version value in prod this way. This PR just pins a static value. We'll need to manually update it when necessary (annoying).
* Add a lil `[dev]` flag on the front-facing version value when not in prod. Obviously this stuff is going to get moved and rearranged anyway but I thought it might be helpful.

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "metakb"
-version = "2.0.0-dev2"
+version = "2.0.0-a1"
 authors = [
     {name = "Alex H Wagner"},
     {name = "Kori Kuzma"},


mcannon068nw:
> > I'm getting unknown still: Screenshot 2025-10-23 at 9 37 08 AM
> > My guess is that this is happening bc I am not running it in a prod environment?
> > (Basing this on this code snippet:
> > ```
> > <Typography>
> >         {version ? `version ${version}` : 'version ?'}
> >         {environment && environment != 'prod' ? ` [${environment}]` : ''}
> > </Typography>
> > ```
> 
> Are you pointing to the dev instance for the api? I'm betting that's why!

I think so!

My vite config is still set to:

```
export default defineConfig({
  plugins: [react()],
  server: {
    proxy: {
      '/api': {
        target: 'http://metakb-dev-eb.us-east-2.elasticbeanstalk.com/',
        changeOrigin: true,
        secure: true,
      },
    },
  },
})
```

mcannon068nw (approved):

PR #627, opened 2025-10-31
docs: instructions for running frontend

### LINKED ISSUE #625: Add documentation for setting up local frontend
I'm not sure if this exists somewhere else, but the README lacks commands that users have to run to start their local frontend. Similar to what we do with setting up the backend, I think we should have commands that users should run to start their local frontend.

katie-perry:
close #625


korikuzma (approved):
🚀

PR #631, opened 2025-11-07
feat: add visualization for evidence level breakdown by row

katie-perry:
Screenshot 2025-11-07 at 3 32 32 PM


client/src/components/ResultTable/ResultTable.tsx
```diff
@@ -110,7 +114,55 @@ const ResultTable: FC<ResultTableProps> = ({ results, resultType }) => {
...
+            <PieChart width={40} height={40}>
+              <Tooltip />
+              <Pie
+                data={data}
+                dataKey="value"
+                nameKey="name"
+                innerRadius={12}
+                outerRadius={20}
+                paddingAngle={2}
+                label={false}
```
jsstevenson: I think these should load a little quicker (default is 1000)

```suggestion
                label={false}
                animationDuration={200}
```

jsstevenson (changes requested):
1 comment on animation speed +

Screenshot 2025-11-10 at 3 40 02 PM

We might want to think about this popup thing a bit more? It took me a few seconds to realize that a) the hover over happens literally on the circle part, not the whole element area, and b) that it refers to that section of the pie chart. It doesn't help that the hover-over most likely covers up the rest of the figure so you have to hunt a little bit more to get it to change. I think this is probably just a limitation of rechart.

I kinda think it would be nicer to just use the MUI tooltip to make a single popup show the value+counts for all at once but that might just be me. Pardon my very crude effort at an example:

Screenshot 2025-11-10 at 3 55 15 PM

katie-perry:
> 1 comment on animation speed +
> 
> Screenshot 2025-11-10 at 3 40 02 PM
> We might want to think about this popup thing a bit more? It took me a few seconds to realize that a) the hover over happens literally on the circle part, not the whole element area, and b) that it refers to that section of the pie chart. It doesn't help that the hover-over most likely covers up the rest of the figure so you have to hunt a little bit more to get it to change. I think this is probably just a limitation of rechart.
> 
> I kinda think it would be nicer to just use the MUI tooltip to make a single popup show the value+counts for all at once but that might just be me. Pardon my very crude effort at an example:
> 
> Screenshot 2025-11-10 at 3 55 15 PM

I really like this idea! I'll make this update

jsstevenson (dismissed):

jsstevenson (approved):

PR #632, opened 2025-11-10
cicd: integrate `precommit` stuff into husky/lint-staged

### LINKED ISSUE #577: `husky` seems to be breaking `precommit`
If you

1. Install server environment and setup precommit (`pip install -e server[dev]`, `pre-commit install` etc)
2. Install client environment (`pnpm install`)

And then try to commit something in any part of the repo, you'll notice that the server precommit hooks no longer fire, even if you have, say, intentionally introduced trailing whitespace or something of that nature:

```
[ feat/475-deploy-frontend ⚙ .venv] code/metakb/server % git commit -m 'fix things?'

> metakb@ lint-staged /Users/jss009/code/metakb
> lint-staged

✔ Backed up original state in git stash (5e9c2424)
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Cleaning up temporary files...
[feat/475-deploy-frontend 2e615aad] fix things?
 3 files changed, 19 insertions(+), 11 deletions(-)
```

And in fact, trying to reinstall precommit hooks will fail:

```
[ testing-precommit ⚙ .venv] ~/code/metakb % pre-commit install
[ERROR] Cowardly refusing to install hooks with `core.hooksPath` set.
hint: `git config --unset-all core.hooksPath`
```

Running `pre-commit run` still works, but it appears that `husky` overwrites the git repo hooks in a way that removes precommit. This makes sense, they're sort of redundant. But it's unfortunate because it means we lose a lot of precommit hooks that we like for code quality, like trailing whitespace, line ending standardization, etc.

A few ways forward

1. Add the precommit hooks into the husky config. I imagine this is possible with some doing, the docs seem very light on details but I'm sure it's possible
2. Try to add the husky behavior in precommit. I think I know how to do this.
3. Try something like [prek](https://github.com/j178/prek)

jsstevenson:
close #577 

I don't think this streamlines setup or anything, but it does seem to make it so that all hooks always run. I hope.

* bumps `lint-staged` version number
* changes autofix behavior during precommit to match our other projects -- now, if something gets autofixed (eg trailing whitespace removed), that change is *kept* (not rolled back) but is not staged, so you can see what changes were made by hooks, and then accept them and reattempt a commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ MetaKB uses automated tooling to maintain consistent code quality across the pro
 - To manually check or fix:
 
   ```bash
-  ruff check server
+  ruff check --fix server
   ruff format server
   ```
 
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "dev": "turbo run dev --parallel",
     "prepare": "husky",
-    "lint-staged": "lint-staged",
+    "lint-staged": "lint-staged --fail-on-changes --continue-on-error --no-revert",
     "format": "pnpm --filter client format && ruff check server --fix && ruff format server"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -15,8 +15,11 @@
   },
   "lint-staged": {
     "server/**/*.py": [
-      "ruff check --fix",
-      "ruff format"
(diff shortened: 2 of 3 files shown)


katie-perry (approved):
looks good to me!

PR #636, opened 2025-11-13
feat: add footer and news section

katie-perry:
Screenshot 2025-11-13 at 4 53 25 PM

- Added the footer
- Moved the GH link and version to the footer
- Did some styling adjustments
- the page looked really boring and bland even with the footer so I added a lil news section for now to try to break up the monotony of the page as it currently is (lmk if I should do this in a different PR or something)

### DIFF
--- a/client/src/pages/Results/ResultPage.tsx
+++ b/client/src/pages/Results/ResultPage.tsx
@@ -26,6 +26,7 @@ import {
   getEntityMetadataFromProposition,
 } from '../../utils'
 import { EvidenceLegend } from '../../components/ResultTable/EvidenceLegend'
+import Footer from '../../components/Footer'
 
 type SearchType = 'gene' | 'variation'
 const API_BASE = '/api/search/statements'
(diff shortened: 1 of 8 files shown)


katie-perry:
image
why does it still look so boring

I need to go to UI design school

client/src/pages/Results/ResultPage.tsx
```diff
@@ -250,9 +251,9 @@ const ResultPage = () => {
   const hasFilteredResults = filteredResults.length > 0
 
   return (
-    <>
+    <Box sx={{ minHeight: '100vh', display: 'flex', flexDirection: 'column' }}>
       <Header />
```
katie-perry: I'm torn about the fact that we specify this on every page and will need to do this for future pages. I opted to do it this way because there was an issue I was working on one time for a different project where I didn't want the same "100vh container, header, footer" pattern. But should I make it so we have to repeat this everywhere on every new page instead of a possible one-off for a page where we may not want it?
🤔 
@jsstevenson do you have any thoughts?

katie-perry: yeah after doing some more research and talking it out with chattyg, I'm going to move this into a layout container of some sort. I'll just go ahead and change that in this PR

jsstevenson:
> why does it still look so boring

looks fine imo, but it will look better as we add more stuff! Do we have a discussion open for what should go in the footer?

jsstevenson (approved):

PR #651, opened 2025-11-27
feat: add '<gene> mutation' catvars from MOA

jsstevenson:
* Identify MOA variants that can be represented with [FeatureContextConstraints](https://cat-vrs.ga4gh.org/en/latest/concepts/catvrs_model.html#featurecontextconstraint) and construct said constraints. I think this could account for up to 200 additional statements.
* Update DB ingest method to block upload of unsupported catvars on the DB side. Once this PR is approved, a subsequent ticket could handle ingest and retrieval of them.

MOA examples: https://moalmanac.org/assertion/559, https://moalmanac.org/assertion/521

See here for an example of CDM output:

https://gist.github.com/jsstevenson/5fc111fdff9e7a23efba18bf682341bc

### DIFF
--- a/server/src/metakb/services/manage_data.py
+++ b/server/src/metakb/services/manage_data.py
@@ -98,20 +98,24 @@ def is_loadable_statement(statement: Statement) -> bool:
             proposition.subjectVariant,
         )
         success = False
-    if len(proposition.subjectVariant.constraints) != 1:
-        _logger.info(
-            "%s could not be loaded because it contains more than 1 constraint: %s",
-            statement.id,
-            proposition.subjectVariant.constraints,
-        )
-        success = False
-    if proposition.subjectVariant.constraints[0].type != "DefiningAlleleConstraint":
-        _logger.info(
-            "%s could not be loaded because it doesn't use a DefiningAlleleConstraint: %s",
-            statement.id,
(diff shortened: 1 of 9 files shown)


DanielPuthawala (approved):
The Cat-VRS portion of the example looks correct to me!

server/src/metakb/services/manage_data.py
```diff
@@ -98,6 +98,24 @@ def is_loadable_statement(statement: Statement) -> bool:
...
         )
         success = False
+    else:
+        if len(proposition.subjectVariant.constraints) != 1:
+            _logger.info(
+                "%s could not be loaded because it contains more than 1 constraint: %s",
+                statement.id,
+                proposition.subjectVariant.constraints,
+            )
+            success = False
```
mcannon068nw, line 108: will there ever be a case where a subject variant has more than one constraint?

mcannon068nw (approved):

server/src/metakb/services/manage_data.py
jsstevenson: Yep! This is a planned future thing but isn't super possible currently. A lot of the parts of this code are basically designed as guards against stuff that isn't supported yet but will be supported eventually

PR #653, opened 2025-11-28
feat: ingest FeatureContextConstraint-based catvars

jsstevenson:
* Enable DB ingest of CDM files containing FeatureContextConstraint-based catvars (this is not possible yet on the current staging branch but can be produced by the code in #651)
* Enable retrieval of statements with FeatureContextConstraint-based subjects by the non-variant properties. EG, if a statement proposition consists of (crudely) `{subjectVariation="IDH1 mutation", geneContext="IDH1", objectCondition="low-grade glioma"}`, then searching by `gene=IDH1` or `disease=low-grade glioma` would be able to return this statement along with any other relevant statements. Note that this does NOT implement retrieval by variation -- I've made a [separate discussion](https://github.com/cancervariants/metakb/discussions/654) to work out desired behavior there.
* As far as I can tell, this... just works on the frontend already, since this doesn't change the schema (just uses some more of the existing schema).  eg 

Screenshot 2025-12-01 at 2 19 49 PM


server/tests/unit/services/test_manage_data.py
```diff
@@ -21,6 +21,7 @@ def test_is_loadable_statement(
     assert is_loadable_statement(Statement(**civic_aid6_statement))
     assert is_loadable_statement(statements["civic.eid:7157"])
     assert is_loadable_statement(statements["moa.assertion:66"])
+    assert is_loadable_statement(statements["moa.assertion:120"])
```
katie-perry, line 24: Does this need to be duplicated between the statements json and the test fixture?

katie-perry (changes requested):

server/tests/unit/services/test_manage_data.py
jsstevenson: @katie-perry update: there's a reason here, basically the statements JSON is a bit easier to manage and I think I'd like to slowly move the other test modules to that format in the future, but not ready to do so yet so for now this assertion lives in both. BUT I did find another place that I had duplicated it and I got rid of that.

katie-perry (approved):

Issue #646, opened 2025-11-21
Map evidence codes from VICC mapping to Level A/B/C/D


mcannon068nw:
### Feature description

VICC mappings should be mapped to Level A/B/C/D/E for display in metakb. 

NOTE: We might consider having this done on the back end stored on a node somewhere

### Use case

I'm always frustrated when I go to look at my evidence strength in MetaKB and see Level A displayed alongside <just> A

### Acceptance Criteria

Codes should be sufficiently mapped as to have all VICC evidence correspond to the appropriate Level A/B/C/D/E. The metakb evidence strength displayed in the front end should be one of these enums

### Contribution

None

korikuzma:
@katie-perry has this mapping work on the frontend that we will be migrating to the backend

mcannon068nw:
Discuss Dec 5th @ Requirements

mcannon068nw:
update: this should exist at the AMP/ASCO/CAP aggregate level, not the individual record level

jsstevenson:
The aggregation stuff is probably going to lead us into "epic" territory. We might want to talk through the subcomponents of this task

One hurdle is that the current search is poorly optimized for handling statements that contain other evidence items. IIRC, it makes a new query for every new subset of statements. Instead, we should try to do a single query per level of depth in the overall tree.

katie-perry:
No E in AMP/ASCO/CAP - this only exists in CIViC at the moment

jsstevenson:
The "aggregate" object is a statement, and it contains evidence lines that link out into the statements that it contains. Those statements have genes, diseases, etc with properties taken from the source (ie MOA usually just has a gene name, for their variants they have a few other things like consequence, etc). Currently, we inject some stuff from the normalizers into those entities. Maybe what this allows us to do instead is just position the normalized entities at the level of the aggregate statement instead.

katie-perry:
https://lucid.app/lucidchart/8b231bc8-f627-4997-9d9e-af59903ce348/edit?viewport_loc=1323%2C3042%2C3656%2C1901%2CI_ORDsud6sfD&invitationId=inv_0887b2e5-b72c-4dd8-8e63-3d947d70ee4f
^ Here's how we do supporting evidence "types" (including cohort sizes) in VarCat to determining the default evidence code for the evidence line (which is the aggregate row). I think we can probably reuse at least some of this logic when implementing something similar in MetaKB

[...]

PR #658, opened 2025-12-02
feat: break normalized drug/disease/gene into separate DB nodes

jsstevenson:
### NOTE: this is back in the refrigerator pending some changes posed in #646 

Screenshot 2025-12-08 at 2 08 04 PM

(ignore the stuff in the bottom right, just look at the circles)

This is more of a preparatory backend refactor. It does *not* change any external behavior, only alters how information is stored internally.

Basically, previously, gene/therapy/disease nodes from a source like CIViC had a property, `normalized_id`, that would contain the normalized ID of that node. So a CIViC gene entry might have ID `"civic.gid:999"`, name `"BRAF"`, and normalized_id `"hgnc:1097"`. That was fine if we only needed to perform direct ID lookups against the normalized record (you would basically ask to fetch all nodes where `normalized_id="hgnc:1097"`, but if we wanted to locate any other data at that level (rather than on the source representations) then we'd need to do something else)

This PR instead makes a common, normalized node that source nodes point to (purple=normalized node, red=source nodes):

Screenshot 2025-12-01 at 10 23 04 PM

This enables some nice stuff in the future, like
* Providing data from our normalizers, like aliases/trade names
* Linking up location/accession info for genes
* Providing ontology/hierarchy for disease terms

(This PR also adds a test case to the repository class just to make me feel a little better)


jsstevenson:
I think I am closing this. We are probably going to do something slightly differently for normalized nodes (exciting!)

PR #661, opened 2025-12-09
feat: add heatmap of top 10 variants/diseases (issue-648)

### LINKED ISSUE #648: Heatmap visualization for top variants on results page
### Feature description

The summary statistic visualizations panel on the results page should contain a heatmap that compares the top 10 variants on one axis by the associated disease on the other axis.

Top 10 variants should be the first 10 variants (by current sorting method in the table, taking into account the current filters selected).

### Use case

I'm always frustrated when I want to see disease associations for variant evidence (i need to do it quick in a visualization cause I'm a busy clinical scientist) but the graph has too much inconsequential noise!!

### Acceptance Criteria

There should be a heatmap at the top of the aggregate results table that displays variants vs associated disease for the top 10 unique variants in the table. It should take up 2/3 of the table and display data based off the currently selected filter set from the page

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

katie-perry:
close #648 

This absolutely wrecked my brain

Screenshot 2025-12-09 at 3 40 18 PM

### DIFF
--- a/client/src/pages/Results/VariantDiseaseHeatmap.tsx
+++ b/client/src/pages/Results/VariantDiseaseHeatmap.tsx
@@ -1,0 +1,126 @@
+import { scaleLinear, scaleBand } from '@visx/scale'
+import { HeatmapRect } from '@visx/heatmap'
+import { Group } from '@visx/group'
+import { AxisBottom, AxisLeft } from '@visx/axis'
+import { useTooltip, Tooltip, defaultStyles as tooltipStyles } from '@visx/tooltip'
+
+import { buildVariantDiseaseMatrix, type NormalizedResult } from '../../utils'
+
+interface Props {
+  data: NormalizedResult[]
+  width?: number
+  height?: number
+}
+
+export function VariantDiseaseHeatmap({ data, width = 1000, height = 700 }: Props) {
+  const { rows, variants, diseases } = buildVariantDiseaseMatrix(data)
+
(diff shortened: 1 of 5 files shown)


client/src/pages/Results/VariantDiseaseHeatmap.tsx
```diff
@@ -0,0 +1,147 @@
...
+              textAnchor: 'end',
+            })}
+          />
+        </Group>
+      </svg>
+
+      {tooltipData ? (
+        <Tooltip top={tooltipTop} left={tooltipLeft} style={tooltipStyles}>
+          <strong>{tooltipData.variant}</strong> → <strong>{tooltipData.disease}</strong>
+          <div>{tooltipData.value} supporting evidence</div>
```
jsstevenson: I think this should be just "evidence item(s)", "item(s)", or ["items" if value > 1 else "item"]

jsstevenson (changes requested):
1 small change but looks great otherwise!

one other thing: this should be addressed in the future (I think it's fine for now), but funny things happen if you don't have a lot of results

Image

client/src/pages/Results/VariantDiseaseHeatmap.tsx
katie-perry: oh yeah that's a great idea!

jsstevenson (approved):

PR #662, opened 2025-12-09
cicd: fix husky/lint-staged?

jsstevenson:
When I updated some of the precommit/cicd stuff, I think I accidentally used a bad version of the `lint-staged` library. This PR updates the dependency. After these changes, running `pnpm install` fixed the error (about an unrecognized `lint-staged` command) that i was getting from the precommit hook. But who knows if this is actually a durable fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
   "devDependencies": {
     "husky": "^9.1.7",
     "json-schema-to-typescript": "^15.0.4",
-    "lint-staged": "^15.5.1",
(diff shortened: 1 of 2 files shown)


katie-perry (approved):
omg you are a saint. Thank you for fixing this! 

I tested this by trying to do everything fresh, so I unset + reset precommit and did a fresh pnpm install. I added some garbage code to both the client and server code and it correctly flagged each individually and altogether! 

Image

PR #672, opened 2025-12-29
feat!: group statements in transformers

### LINKED ISSUE #666: Restructure CDMs to reflect aggregated evidence structure
So, it's kind of arbitrary whether this happens in the transformer/CDM output, or in the DB load, or maybe in between. I'm going with the former for now but might revise that. 

Basically the idea is to make every incoming statement be nested as an evidence line inside of an "aggregated" statement. The entities in the proposition of the aggregated statement will be the values gathered from the variation normalizer. The entities in the proposition of the statement itself will just be whatever the source gives us.

Considerations
* This will make the CDMs way bigger. Maybe we should start using iriReferences more.
* Not sure what the IDs of the aggregated statements will be
* Still not totally clear how CIViC assertions are supposed to be handled -- they are "related" to these grouped propositions? or something?
* This will make it a lot easier to bring in properties from the normalizer. We should keep it small at first.

jsstevenson:
close #666


mcannon068nw:
Discuss with @ahwagner at Friday time

PR #673, opened 2026-01-02
feat: update db + api (cdm-updates)

korikuzma:
close #563 

* Updates requirements
* Therapy group relation changed from `HAS_SUBSTITUTE|HAS_COMPONENT` -> `HAS_THERAPY`

## Notes

* You will want to update your dependencies `uv sync --all-extras --reinstall`
* For CDMs, I used pickle file from `server/tests/data/civicpy_cache_20250929.pkl` 

## Comments

~~* Includes `node_type` property on DB Condition (Disease, Phenotype, ConditionSet) Nodes to be used in loading condition set cypher query. Ideally this would be done using labels, but not sure it's quite possible? Might be something to look into more in the future?~~
* Uses 2025-12-29 normalizer ddb images

## Plans

Once this is merged, I'll create a PR to merge `cdm-updates` into `staging` (will do squash & merge)

PR #675, opened 2026-01-02
fix(build): temp pin agct

korikuzma:
Once cdm-updates gets merged (see #673 where I update deps), we'll be able to remove this


jsstevenson:
@korikuzma do we need to update the pipfile too for cases like this?

korikuzma:
> @korikuzma do we need to update the pipfile too for cases like this?

MetaKB uses a requirements file instead of pipfile (I don't see a Pipfile in the codebase)

jsstevenson (approved):

PR #678, opened 2026-01-14
feat!: remove db url construction from secrets

### LINKED ISSUE #643: Remove `METAKB_DB_SECRET`
### Feature description

Instead of storing separate fields in a secret, we should just set these using the `METAKB_DB_URL`

### Use case

* We now have extra code that is not needed

### Acceptance Criteria

Given I want to connect to the AWS neo4j DB,
When I set `METAKB_DB_URL` with the correct credentials,
Then I should be able to connect without any issues

Given I want to connect to the AWS neo4j DB,
When I set `METAKB_DB_SECRET` with the correct credentials,
Then I should NOT be able to connect

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

we'll need to update the secret 

### Contribution

None

katie-perry:
close #643 

I hope this works! 

In this PR, I removed code that gets the db url from the secrets. Instead, we will use the env variable `METAKB_DB_URL` is provided by the service catalog

### DIFF
--- a/server/src/metakb/repository/neo4j_repository.py
+++ b/server/src/metakb/repository/neo4j_repository.py
@@ -1,13 +1,9 @@
 """Neo4j implementation of the repository abstraction."""
 
-import ast
 import logging
-import os
 from typing import NamedTuple
 from urllib.parse import urlparse, urlunparse
 
-import boto3
-from botocore.exceptions import ClientError
 from ga4gh.cat_vrs.models import CategoricalVariant
 from ga4gh.core.models import MappableConcept
 from ga4gh.va_spec.aac_2017 import (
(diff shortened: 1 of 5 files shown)


server/src/metakb/repository/neo4j_repository.py
```diff
@@ -157,18 +134,27 @@ def get_driver(
...
-        secret = ast.literal_eval(_get_secret())
-        url = f"bolt://{secret['username']}:{secret['password']}@{secret['host']}:{secret['port']}"
-    elif url:
-        pass  # use argument if given
+        else:
+            _logger.error(
+                "No DB connection URL provided in %s environment; "
+                "METAKB_DB_URL is expected to be set",
+                configs.env,
+            )
```
jsstevenson, line 147: we could consider just raising a `RunTimeError` to this effect here -- if we're in one of the beanstalk envs, might be nicer to see this as the explicit reason why an instance fails to launch instead of a DB connection error that happens later

katie-perry: I was really struggling to figure out how to structure the code because I wanted to maintain the previously existing funcionality/logging but change this to not rely on the secrets. Ultimately I think we'll end up getting both the logging error and we'll get the Neo4jCredentialsError (line 156) - it just executes after the deployed-environment-specific checks

jsstevenson: yeah that made more sense once I kept reading

jsstevenson (dismissed):

korikuzma (dismissed):

jsstevenson (approved):

PR #692, opened 2026-01-19
Add cBioPortal harvester and transformer infrastructure

hmcostello:
cBioPortal data harvest and transform infrastructure. 

It takes a long time to run though all studies because of the var normalization (10-12 h). You can test code using only one of the studies. 

Run one study from harvester dir:

```python
from metakb.harvesters.cbioportal import cBioportalHarvester
from metakb.transformers.cbioportal.base import cBioportalTransformerBase
harvester = cBioportalHarvester()
data = harvester.harvest("all_stjude_2015")
base = cBioportalTransformerBase()
df = base.run_transformers({"all_stjude_2015": data})
```

To run all studies, just remove "all_stjude_2015" from commands.

@mcannon068nw @korikuzma @jsstevenson @katie-perry


korikuzma (commented):
I think @jsstevenson and/or @katie-perry will be reviewing this (going to add them as specific reviewers instead of the maintainer team)

server/src/metakb/transformers/cbioportal/__init__.py
mcannon068nw, line 1: empty?

are there any imports that should go here?

server/src/metakb/transformers/cbioportal/base.py
```diff
@@ -0,0 +1,886 @@
...
+import time
+from os import environ
+from pathlib import Path
+from typing import Any
+
+import requests
+
+environ["AWS_ACCESS_KEY_ID"] = "dummy"
+environ["AWS_SECRET_ACCESS_KEY"] = "dummy"
+environ["AWS_SESSION_TOKEN"] = "dummy"
```
mcannon068nw, line 18: seeing these a lot in the transformer files, what are these used for? if it's set to dummy it makes me think they aren't doing anything, otherwise these should be set to an os.getenv(key) type deal

server/src/metakb/transformers/cbioportal/transformer_all_phase2_target_2018_pub.py
```diff
@@ -0,0 +1,249 @@
...
+    "TMB_NONSYNONYMOUS",
+]
+
+PATTERN = re.compile(r"^23-")
+
+
+class cBioportalTransformer(Transformer):
+    """A class for transforming cBioportal Data to the common data model."""
+
+    # TODO: TypeError: Can't instantiate abstract class cBioportalTransformer without an implementation for abstract method '_create_cache'
```
mcannon068nw: This seems important to have and could potentially speed up run time. Could the cBioportalTransformerBase have a _create_cache method that all these specific transformers inherit?

server/src/metakb/transformers/cbioportal/transformer_all_stjude_2015.py
```diff
@@ -0,0 +1,264 @@
...
+    ):
+        return super()._get_therapeutic_substitute_group(
+            therapeutic_sub_group_id, therapies, therapy_interaction_type
+        )
+
+    def _get_therapy(self, therapy):
+        return super()._get_therapy(therapy)
+
+    def _create_cache(self):
+        pass
```
mcannon068nw: @jsstevenson These are required imports but don't do anything. The cache seems like it could be something good and I don't think these cBioportal studies have any therapeutic data. Is it possible to loosen required abstract method requirements for situations like this otherwise all the transformers will just have these here even though they aren't doing anything?

server/src/metakb/transformers/cbioportal/transformer_es_iocurie_2014.py
```diff
@@ -0,0 +1,271 @@
...
+        self.samples = pd.DataFrame(harvested_data.samples).filter(SAMPLE_HEADERS)
+        self.metadata = pd.DataFrame(harvested_data.metadata)
+
+        # MUT DF
+        # Strip whitespace and rename col
+        self.variants.columns = self.variants.columns.str.strip()
+        self.variants = self.variants.rename(
+            columns={"Tumor_Sample_Barcode": "SAMPLE_ID"}
+        )
+        self.variants = self.variants.rename(columns={"AAChange": "Amino_Acid_Change"})
```
mcannon068nw: I haven't compared too many of these files with each other but I wonder if there aren't more common methods that can be migrated to the transformer base.

server/src/metakb/transformers/cbioportal/__init__.py
hmcostello: Is that where imports should go?  I added it because one of the errors after creating the PR said I needed one.

server/src/metakb/transformers/cbioportal/base.py
hmcostello: The code wasn't working without these. I can't remember the exact issue, but it was giving me an AWS SSO error for something

mcannon068nw:
Took a cursory look. Some initial impressions are that there is a lot of commented out code for functions that were used in some studies but not in others that we could probably remove for clean-up. Also I wonder if there are some more general methods present in each study that could be abstracted a little bit and moved to base to deduplicate where possible.

server/src/metakb/transformers/cbioportal/transformer_all_phase2_target_2018_pub.py
hmcostello: I suppose it could, but only one study-specific transformer calls the normalizer (for the chromosome 23 issue.)

The rest of the long run-time is from variant normalization, which all happens in the base transformer.

server/src/metakb/transformers/cbioportal/transformer_es_iocurie_2014.py
hmcostello: That's probably true, I can work on that.

hmcostello:
> Took a cursory look. Some initial impressions are that there is a lot of commented out code for functions that were used in some studies but not in others that we could probably remove for clean-up. Also I wonder if there are some more general methods present in each study that could be abstracted a little bit and moved to base to deduplicate where possible.

Yes, I will work on this.

server/src/metakb/transformers/cbioportal/__init__.py
mcannon068nw: tbh not sure how this file works I'll defer to the devs but I peeked at one of the other transformer sets and it has some database specific imports going on:

see: https://github.com/cancervariants/metakb/blob/main/server/src/metakb/transformers/__init__.py

I'm not sure what these are for and if there's anything here on the cBioPortal side that should be mirrored

server/src/metakb/transformers/cbioportal/transformer_es_iocurie_2014.py
mcannon068nw: from code review, just remembered our conversation on data provenance for processing

server/src/metakb/transformers/cbioportal/__init__.py
zealws: There's two use-cases for `__init__.py` files:
1. Telling the python interpreter to treat a directory as a module
2. Exporting a symbol defined elsewhere to be importable from the directory's module, rather than from the file's module.

Suppose I have this structure:
```
.
├── lib
│   └── lib2.py
├── lib1.py
└── main.py
```

If my `main.py` has `from lib1 import helper_fn`, that'll work fine. Python files are modules, and you can import the `lib1` module from the `main` module since they're in the same directory.

If my `main.py` has `from lib.lib2 import helper_fn2` that won't work. Directories are not python modules (by default) so you can't import anything from `lib`.

You can tell the python interpreter that `lib/` is a python module by creating an `__init__.py` file within that directory: `touch lib/__init__.py`.

Now `from lib.lib2 import helper_fn2` will work fine.

You could stop there, leave `__init__.py` empty, and use whatever you need in the directory, and it'll work fine.

But what if you wanted to bring a helper function from `lib.lib2` "up" into the `lib` module?

E.g. to do `from lib import helper_fn2` from `main.py`?

`helper_fn2` isn't defined in the lib's `__init__.py`, it's in `lib/lib2.py`. But we can modify `__init__.py` to re-export `helper_fn2` so it'd be importable like this.

You'd add `from .lib2 import helper_fn2` in `__init__.py`, and then you could import `from lib import helper_fn2` from `main.py` and it'd work.

This is the reason why `__init__.py` statements often have imports. It's to re-export some nested object/variable/class/etc at the level of the directory's module, rather than at the level of the file.

Also note that many python linters don't like imports that aren't used. You should only import code that file is actually going to use. But the `__init__.py` is just importing stuff so it can be imported elsewhere, so it's not going to be used within that file. That's why `# noqa: F401` is on those lines in the one Matt linked. It disables the [lint error for unused import](https://www.flake8rules.com/rules/F401.html).

server/src/metakb/transformers/cbioportal/base.py
zealws: I think what @mcannon068nw's suggesting is using something like:

```py
environ["AWS_ACCESS_KEY_ID"] = os.getenv("AWS_ACCESS_KEY_ID", "dummy")
```

This will not replace `AWS_ACCESS_KEY_ID` if it's already set, but if it's not set, it'll set it to `dummy`.

This would let someone set `AWS_ACCESS_KEY_ID` if they wanted.

server/src/metakb/transformers/cbioportal/transformer_all_stjude_2015.py
zealws: > Is it possible to loosen required abstract method requirements for situations like this otherwise all the transformers will just have these here even though they aren't doing anything?

Not without modifying `Transformer`. The only ways to suppress that error are to define empty stub methods like what's in the code already, or modify `Transformer` to put the stub there.

Alternatively, you could define a `Transformer` subclass like `UncachedTransformer` that stubs out the method, and have things inherit from that if they don't need caching.

jsstevenson (dismissed):

mcannon068nw:
Leaving open to preserve comments and code until all transformers are through. NOTE: Heather feel free to close early if you don't need this

mcannon068nw:
This work has been completed in 16 / 17 other PRs

PR #695, opened 2026-02-03
feat: add cbioportal harvester

### LINKED ISSUE #437: Write Harvester for cBioPortal
cBioPortal contains a wealth of cancer variant datasets that we are not incorporating in any form. Specifically, we are interested in pulling in Pediatric Cancer Studies data and using this to support Cohort Allele Frequencies. 

We should write a data harvester that ingests and deduplicates this dataset for MetaKB users.

Side note, data that we pull from this source will likely be affected by discussion on what is necessary for Cohort Allele Frequency.

 https://www.cbioportal.org/

jsstevenson:
close #437 

Add cBioPortal Harvester
* Implement a wags-tails data fetcher for the designated cBioPortal studies. Wags-tails isn't really designed for a) multiple distinct datasets under one name and b) multiple data files per dataset (wish we had thought of this when we first wrote it) so the API technically breaks a bit, but everything functions properly here. We could eventually produce a more general solution for wags-tails and move it over there, but I think this is ok for now
* Use pandas to convert text files into dataframes and then dump into JSON stored under the wags-tails `metakb/cbioportal/harvester/` data directory.
* Update CLI to enable `metakb harvest cbioportal` command.
* Add simple regression test

Collab w/ @hmcostello


mcannon068nw (approved):

server/src/metakb/harvesters/cbioportal.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self._download_data(file_paths)
+        return file_paths, ""
+
+
+class CBioPortalHarvestedData(_HarvestedData):
+    """Provide harvested data for a cBioPortal study"""
+
+    study_name: CBioPortalStudyName
+    variants: list[dict]
```
katie-perry, line 148: don't have to address in this PR but it would be great to create a followup techdebt ticket to make these types or Pydantic objects for what the individual objects look like, if possible!

server/tests/data/cbioportal/cbioportal_all_stjude_2015_data_clinical_sample.txt
```diff
@@ -0,0 +1,10 @@
+#Sample Identifier	Patient Identifier	MLL Status	Leukemia Type	Mutation Rate	Gene Panel	Cancer Type	Cancer Type Detailed	Oncotree Code	Somatic Status	TMB (nonsynonymous)
+#A unique sample identifier.	Identifier to uniquely specify a patient.	MLL status.	Leukemia Type	Mutation Rate	Gene panel.	Disease type	Cancer Type Detailed	Oncotree Code	Somatic Status.	TMB (nonsynonymous)
+#STRING	STRING	STRING	STRING	NUMBER	STRING	STRING	STRING	STRING	STRING	NUMBER
+#1	1	1	1	1	1	1	1	1	1	1
+SAMPLE_ID	PATIENT_ID	MLL_STATUS	LEUKEMIA_TYPE	MUTATION_RATE	GENE_PANEL	CANCER_TYPE	CANCER_TYPE_DETAILED	ONCOTREE_CODE	SOMATIC_STATUS	TMB_NONSYNONYMOUS
+SJINF001	SJINF001	INF_MLLr	Pre-B	3.97E-08	WGS	Leukemia	B-Cell Acute Lymphoid Leukemia	BALL	Matched	0
```
katie-perry, line 6: Are these real data? Do we need to worry about modifying these or doing any sort of further de-identification?

jsstevenson: it's public data

server/src/metakb/schemas/app.py
```diff
@@ -20,7 +21,9 @@ def as_print_case(self) -> str:
             return "CIViC"
         if self == "moa":
             return "MOA"
-        raise ValueError
+        if self == "cbioportal":
```
katie-perry, line 24: it looks like you're just following the established pattern here, but why aren't we using the constants above in these checks? We could prob refactor this
(just noting - this is another nitpick that I don't mind if we address later in a different PR)

katie-perry (approved):

server/src/metakb/schemas/app.py
jsstevenson: 👍 I'm gonna try to remember to grab this next PR

Issue #691, opened 2026-01-16
Update civicpy to 5.1.1


korikuzma:
### Feature description

In #690 , we have civicpy pinned to a specific commit (d55d0a5d4e96). However, we should really be using the latest pypi version (5.1.1).

https://github.com/griffithlab/civicpy/compare/d55d0a5d4e96...5.1.1

### Use case

* Good to use pypi version instead of git reference
* Pull in @katie-perry 's changes for capturing approval info

### Acceptance Criteria

Given I want to know which version of civicpy I'm using,
When I run `uv pip show civicpy` or `pip show civicpy` in my metakb venv,
Then I should get 

```
Name: civicpy
Version: 5.1.1
```

Given I want to confirm tests still work,
When I run `pytest`,
Then all tests should pass
Note: will need to replace `server/tests/data/civicpy_cache_20250929.pkl` with updated cache (`civicpy update --hard`, then move and rename with the date this was generated). We should only expect minor changes (molecular profile score and `is_approved_vcep` extension in `Contribution`s from @katie-perry's work)

### Contribution

None

PR #697, opened 2026-02-11
build!: update civicpy to latest version

katie-perry:
#691 

opening as draft for now as I do final testing


jsstevenson:
I think we should check whether upgrading to the latest release of sphinxcontrib-images fixes this

jsstevenson (approved):

PR #701, opened 2026-02-11
build: update civicpy to 5.1.1

katie-perry:


requirements.txt
```diff
@@ -1,117 +1,113 @@
-# This file was autogenerated by uv via the following command:
-#    uv pip compile pyproject.toml --extra deploy -o ../requirements.txt --no-annotate
```
katie-perry, line 2: wait. dangit, the README said to do `pip freeze --exclude-editable > ../requirements.txt`
which am I supposed to do?

katie-perry: disregard- ran this command instead. I'll update the readme

jsstevenson: Yeah it should be exactly the same I think? minus the note at the top

jsstevenson (approved):

jsstevenson (approved):

PR #702, opened 2026-02-11
fix(ci): add unknown fallback_version

korikuzma:
In cases where setuptools-scm cannot determine a version due to missing git metadata, set a default unknown fallback version so builds don't fail.


jsstevenson:
> setuptools-scm

This is not how the version is set, right? I don't think we are using setuptools-scm for metakb

korikuzma:
> > setuptools-scm
> 
> This is not how the version is set, right? I don't think we are using setuptools-scm for metakb

Ope, apologies. I forgot it was removed from MetaKB (we had to do this for AnyVLM to get the build to work)

PR #708, opened 2026-02-16
feat: add tqdm to transformers

jsstevenson:
Add `tqdm` (progress bar) to transformers. (connected to #526 but doesn't complete it)

This isn't necessary per se, but it does help with debugging/peace of mind. It also coincides with a small refactor to write functions that operate on a per-entity bases (gene, variant) rather than functions that transform the entire set of those entities at once (which is most of the code changes here). Otherwise, there are no behavior changes.

the tqdm pbar is manually incremented because the transformers are looping through many different kinds of entities depending on the source. This is a pattern that we will move away from in subsequent PRs, but I'm trying to minimize the changes here.

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -46,6 +46,7 @@ dependencies = [
     "asyncclick",
     "jinja2",
     "pandas>=2.3.3, <3.0",
+    "tqdm",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


korikuzma:
Related discussion: https://github.com/cancervariants/metakb/discussions/526

mcannon068nw (approved):

PR #709, opened 2026-02-16
feat!: use cross-source ID generation logic

jsstevenson:
We have to mint custom IDs for concept sets in some sources. We will have this problem for every normalized concept set at the level of aggregated evidence, so we want to pull this logic out to a general place and write it once instead of making decisions on a per-source basis. It looks like this:

```pycon
>>> self._compute_combo_id(
...     "MOA",
...     TherapyGroup,
...     MembershipOperator.AND,
...     ["moa.therapy:imatinib", "moa.therapy.trastuzumab"],
... )
'MOA.tg:ojf-glrsMg7fNrGKoGwGEF0OTyssCuCA'
>>> _compute_combo_id("metakb", TherapyGroup, MembershipOperator.AND, ["rxcui:2555", "ncit:C1926"])
'metakb.tg:mBrVr9O_7ZbfDyBdauw3uXF3bXnmApGg'
```

I could be convinced that this should just be `"therapeutic"` and `"condition"` instead of `"tg"` and `"cs"` (for TherapyGroup and ConditionSet).


server/src/metakb/transformers/base.py
```diff
@@ -300,13 +301,60 @@ def __init__(
...
+    ) -> str:
+        """Compute identifier for therapy group or condition set
+
+        >>> self._compute_combo_id(
+        ...     "MOA",
+        ...     TherapyGroup,
+        ...     MembershipOperator.AND,
+        ...     ["moa.therapy:imatinib", "moa.therapy.trastuzumab"],
+        ... )
+        'MOA.tg:ojf-glrsMg7fNrGKoGwGEF0OTyssCuCA'
```
korikuzma:
```suggestion
        'moa.tg:ojf-glrsMg7fNrGKoGwGEF0OTyssCuCA'
```

server/src/metakb/transformers/base.py
```diff
@@ -300,13 +301,60 @@ def __init__(
...
+
+        These values should generally be for internal use only, so it's not super
+        important that they are especially meaningful, but they should be consistent
+
+        :param source_prefix: prefix to use in ID namespace
+        :param combo_class: type of entity combination
+        :param operator: generally either "AND" or "OR"
```
korikuzma: I don't think this will ever change
```suggestion
        :param operator: Either "AND" or "OR"
```

server/src/metakb/transformers/base.py
```diff
@@ -300,13 +301,60 @@ def __init__(
...
+        :param ids: list of entity IDs to combine
+        :return: CURIE designating the combination in a deterministic way
+        """
+        if not all(ids):
+            raise ValueError
```
korikuzma, line 344: Add raise ValueError to docstring

server/src/metakb/transformers/base.py
```diff
@@ -300,13 +301,60 @@ def __init__(
...
+        # use the whole membership operator string to distinguish from an improbable clash w/ a real entity name
+        ids.append(str(operator))
+        ids.sort()
+        blob = json.dumps(ids, separators=(",", ":"), sort_keys=True).encode("ascii")
+        digest = sha512t24u(blob)
+        if combo_class is TherapyGroup:
+            combo_class_abbrev = "tg"
```
korikuzma, line 353: Wonder if we should create enums for these?

server/src/metakb/transformers/base.py
```diff
@@ -300,13 +301,60 @@ def __init__(
...
+        elif combo_class is ConditionSet:
+            combo_class_abbrev = "cs"
+        else:
+            raise ValueError
+        return f"{source_prefix.lower()}.{combo_class_abbrev}:{digest}"
+
+    ### Other stuff (TODO: reorganize in future PRs on this branch)
```
korikuzma: Does issue exist for TODO?

korikuzma (changes requested):

server/src/metakb/transformers/base.py
jsstevenson: Left a note with a checkpoint

server/src/metakb/transformers/base.py
jsstevenson: If there are other cases where we expect to be reusing these specific string values, then yes, but otherwise I think no? I thought about defining a dict like

```python
combo_class_prefix_map = {
    ConditionSet = "cs",
    TherapyGroup = "tg",
}
```

but I wasn't sure where else it would get used outside of this conditional

korikuzma (approved):

PR #710, opened 2026-02-17
feat!: move classification methods out of shared space

jsstevenson:
* technically a "feat!" but really just a refactor

Move Method (capital M) construction out of a shared base class space and into individual source classes. The CIViC transformer already does this since civicpy provides the method now, and the constructor function pattern used in the MOA class here will match up with how other parts of the GKS statement get built.

### DIFF
--- a/server/src/metakb/transformers/moa.py
+++ b/server/src/metakb/transformers/moa.py
@@ -85,6 +85,10 @@ class MoaTransformer(Transformer):
         return _MoaTransformedCache()
 
     def _create_method(self) -> Method:
+        """Get MOA classification method object for use in study statements
+
+        :return: MOA method
+        """
         return Method(
             id="moa.method:2021",
             name="MOAlmanac (2021)",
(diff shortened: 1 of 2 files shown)


server/src/metakb/transformers/moa.py
```diff
@@ -78,16 +78,24 @@ def __init__(
...
-        self.processed_data.methods = [
-            self.methods_mapping[MethodId.MOA_ASSERTION_BIORXIV.value]
-        ]
         self._cache = self._create_cache()
 
     def _create_cache(self) -> _MoaTransformedCache:
         """Create cache for transformed records"""
         return _MoaTransformedCache()
 
+    def _create_method(self) -> Method:
```
katie-perry, line 87: I know there wasn't one before but.... 

docstring? 
🥺
 👉 👈

jsstevenson: yup

katie-perry (approved):

Issue #665, opened 2025-12-13
Merge evidence into aggregated nodes upon DB load


jsstevenson:
The repository loader and query functions should be updated to handle the new CDM evidence structure. For the most part, everything should still work, but a few tweaks will be needed

- Remove `normalized_id` from the entity nodes. Don't search on them anymore. Instead, search directly on ID (since the aggregate statement will be linked to a node with a normalized ID as its ID)
- Update the test fixtures to reflect how inputs look now. No more `vicc_normalizer_priority` etc
- Statements can now be three-leveled (metakb statement -> civic assertion -> civic evidence item). Make sure these can be accurately ingested, searched for, and retrieved.

jsstevenson:
closed by #751

Issue #664, opened 2025-12-13
Create higher-order MetaKB assertions composed of normalized source statements


jsstevenson:

jsstevenson:
I'm closing this for now -- there are a couple remnant issues but we can handle them separately

Issue #178, opened 2022-07-14
Include metadata in harvest/cdm files


korikuzma:
Possible data could include (it's ok if an initial ticket doesn't get everything)

- [ ] Timestamp
- [ ] Version of the source if available (e.g. civic version)
- [ ] Version of software used to acquire from the source (e.g. civicpy version)
- [ ] MetaKB version
- [ ] Normalizer versions
- [ ] VRS, VA-Spec, Cat-VRS versions

Then

- [ ] Figure out a way to put this on the DB
- [ ] Figure out a way to get this out of the DB
- [ ] Put it in the search response

jsstevenson:
See also https://github.com/cancervariants/metakb/issues/394

mcannon068nw:
what version of civicpy was used to create the dump, etc

mcannon068nw:
minimum, get the date associated with the file sounds good

PR #714, opened 2026-02-19
feat!: aggregate evidence in transformers

### LINKED ISSUE #666: Restructure CDMs to reflect aggregated evidence structure

jsstevenson:
close #666

* perform evidence aggregation. Rather than modifying incoming evidence items in-place, first try to faithfully construct it as a VA-Spec object, then attempt to normalize each component entity; if successful, the source statement is included in an evidence line.
* refactor the transformer pipeline, generally. I restructured CIVIC and MOA to both work like the following: for each statement, 1) convert to a minimal GKS-compliant object, 2) attempt to normalize each proposition component, 3) if successful, construct the aggregated metakb assertion and include the source statement in an evidence line. 
    * Moved some stuff out to a few new modules, `identifiers.py` and `catvars.py`, that didn't rely on any transformer state and referred to a discrete set of tasks. 
    * Since the transformers are first converting input objects to a standardized GKS structure, you don't need any special knowledge of the source to attempt entity normalization; as a result, these methods can mostly live in the base class now, although variant normalization is still finicky enough that each transformer is supposed to reimplement that itself
    * simplify caching and add configurable LRU caching to the normalizers themselves. This is described more in #711 but I've closed that PR since it's just included in this one.
* remove a lot of tests
     * rewrite transformer tests in #718 
     * Some other tests are still failing because they relied upon now-unused modeling elements as inputs/expected outputs. This should be addressed in #665 

See example output here https://gist.github.com/jsstevenson/80868828189d76db64f318b0f511d3e1

Future investigation:
* Classify(?) MetaKB assertions based on contained evidence; classify individual statements by kind of evidence and sort into appropriate evidence line


mcannon068nw:
ship it

mcannon068nw:
tests as additional module afterwards to unblock evidence agg stuff

server/src/metakb/repository/neo4j_models.py
```diff
@@ -537,16 +537,17 @@ class DrugNode(BaseNode):
...
+        # for mapping in therapy.mappings:
+        #     for ext in mapping.extensions:
+        #         if ext.name == NormalizerExtensionName.PRIORITY and ext.value:
+        #             normalized_id = mapping.coding.id
+        #             break
+        #     if normalized_id:
+        #         break
+        # else:
+        #     msg = f"Unable to locate normalized ID in therapy {therapy}"
+        #     raise ValueError(msg)
```
mcannon068nw: question, #664 is the epic for this, which sub-issue is intended for this block.Is this #646?

server/src/metakb/transformers/base.py
```diff
@@ -3,91 +3,72 @@
...
-
-    :param name: Name to sanitize
-    :return: Sanitized string with whitespace characters replaced by underscores
-    """
-    return re.sub(r"\s+", "_", name.strip())
-
-
-class NormalizerExtensionName(str, Enum):
-    """Define constraints for normalizer extension names"""
+# TODO figure out classification, method, etc for MetaKB assertions
```
mcannon068nw: do we have an issue for this

server/src/metakb/transformers/base.py
```diff
@@ -368,207 +278,362 @@ def extract_harvested_data(self) -> _HarvestedData:
...
+        """Create a composite JSON for transformed data.
 
-        :return: Dictionary containing mapping from source evidence level (key)
-            to corresponding vicc concept vocab (value) represented as a list of
-            ConceptMapping
+        :param cdm_filepath: Path to the JSON file locatio at which the CDM output will be
+            saved. If not provided, will use the default path of
+            ``<METAKB_DATA_DIR>/<src_name>/transformers/<src_name>_cdm_YYYYMMDD.json``,
+            where ``<METAKB_DATA_DIR>`` is the configurable data root directory.
+            See the :ref:`configuration <config-data-directory>` entry in the docs for more information.
```
mcannon068nw, line 288: last year the idea of use-case specific cdms was discussed (e.g. user A only gets public data and user B uses private institutional data or something, or maybe even user C only wants CDMs containing pediatric data for their instance).

this method looks like it saves everything for a single source, is this easily adaptable for a couple of these different use cases?

server/src/metakb/transformers/base.py
```diff
@@ -368,207 +278,362 @@ def extract_harvested_data(self) -> _HarvestedData:
...
+                    members.append(condition)
+                elif condition.conceptType == "Disease":
+                    normalized_disease = self._normalize_disease(condition)
+                    if not normalized_disease:
+                        return None
+                    members.append(normalized_disease)
+                else:
+                    raise ValueError
+            elif isinstance(condition, ConditionSet):
+                normalized_condition_set = self._resolve_condition_set(condition)
```
mcannon068nw, line 372: this was confusing at first but i think I understand this.

Are we expecting conditionSets to contain conditionSets? if so, how frequently does this occur and how deep does it go usually?

server/src/metakb/transformers/base.py
```diff
@@ -368,207 +278,362 @@ def extract_harvested_data(self) -> _HarvestedData:
...
+        :param condition_set: source-derived ConditionSet
+        :return: concept set with normalized equivalents of all inputs, or ``None`` if unsuccessful
+        """
+        members: list[MappableConcept | ConditionSet] = []
+        for condition in condition_set.conditions:
+            if isinstance(condition, MappableConcept):
+                if condition.conceptType == "Phenotype":
+                    members.append(condition)
+                elif condition.conceptType == "Disease":
+                    normalized_disease = self._normalize_disease(condition)
```
mcannon068nw, line 365: Phenotype = age of onset?

server/src/metakb/transformers/base.py
```diff
@@ -368,207 +278,362 @@ def extract_harvested_data(self) -> _HarvestedData:
...
-        mappings: list[ConceptMapping] = []
-        attr_name = NORMALIZER_INSTANCE_TO_ATTR[type(normalizer_resp)]
-        normalizer_resp_obj = getattr(normalizer_resp, attr_name)
-        normalizer_label = normalizer_resp_obj.name
-        is_disease = isinstance(normalizer_resp, NormalizedDisease)
-        is_gene = isinstance(normalizer_resp, NormalizedGene)
-        is_therapy = isinstance(normalizer_resp, NormalizedTherapy)
+        Treat phenotypes as "normalized" and copy them up to the normalized object. In
+        the future, we might want to be more careful about this, maybe check for a
+        preferred namespace or try to map over to HPO.
```
mcannon068nw, line 394: if phenotypes are 'age of onset' type data, do we have any thoughts on a "normalization" step to bucket them according to the same buckets that Alex provided for Heather's cBioPortal work?

mcannon068nw: https://docs.google.com/document/d/17fpD6PbAROW0Qm7bM_bVJ8vexe5FFLP3eNGqcwgOia8/edit?tab=t.0

the green, blue and yellow groupings there

server/src/metakb/transformers/base.py
```diff
@@ -368,207 +278,362 @@ def extract_harvested_data(self) -> _HarvestedData:
...
+        | VariantTherapeuticResponseStudyStatement
+        | None
+    ):
+        if isinstance(statement.proposition, VariantTherapeuticResponseProposition):
+            return await self._build_aggregated_tr_statement(statement)
+        if isinstance(statement.proposition, VariantDiagnosticProposition):
+            return await self._build_aggregated_diag_statement(statement)
+        if isinstance(statement.proposition, VariantPrognosticProposition):
+            return await self._build_aggregated_prog_statement(statement)
+        raise ValueError
```
mcannon068nw, line 519: will a statement will only ever be one of these at a time?

possible place to use a `match` `case` block?

server/src/metakb/transformers/catvars.py
```diff
@@ -0,0 +1,106 @@
...
+                        primaryCoding=Coding(
+                            code=code("is_within"),
+                            system="ga4gh-gks-term:location-match",
+                        )
+                    ),
+                    relations=[LIFTOVER_TO_RELATION],
+                )
+            ),
+        ],
+    )
```
mcannon068nw, line 62: is this placeholder? we aren't displaying any copynumberchange data right now (only really SNVs and indels?) I had thought. If we have it, that rules and we should start thinking about how to display it on the front end.

server/src/metakb/transformers/civic.py
```diff
@@ -726,32 +280,41 @@ def _parse_mp_name(
...
+        parsed_components: MolecularProfileNameComponents,
+        variant_expr: str,
+        variant_id: str,
+    ) -> bool:
+        """Validate whether the variant appears to be normalizable as a known CatVar
+
+        For now, this is a yes/no check. In the future, when more kinds of catvars are
+        supported, I think this should change into a "variant classifier" that returns
+        what kind of variant the name appears to be, and then that can be checked against
+        a list of supported/unsupported variants and dispatched accordingly.
```
mcannon068nw, line 298: This is an interesting idea. Maybe a dumb question and I'm just not familiar enough with CIViC, but do MolecularProfileNames use a "standard enough" vocabulary to support inference of "what kind of variant the name appears to be"

server/src/metakb/transformers/civic.py
```diff
@@ -726,32 +280,41 @@ def _parse_mp_name(
...
         """
-        vname_lower = molecular_profile_name.lower()
-
-        is_frameshift = vname_lower.endswith("fs")
-        has_unsupported_chars = any(c in vname_lower for c in ("-", "/"))
-        unable_to_normalize_names = bool(
-            set(vname_lower.split()) & UNABLE_TO_NORMALIZE_VAR_NAMES
-        )
+        if parsed_components.c_change:
+            msg = f"cDNA variant (ID: {variant_id}) not yet supported. This will be added in issue-225"
```
mcannon068nw, line 306: is this issue urgent enough to push up the priority list? i wonder what the variant add if we fix this will be

server/src/metakb/transformers/identifiers.py
```diff
@@ -0,0 +1,103 @@
...
+    :param str_array: array of string values (e.g. IDs, predicate terms, etc)
+    :return: hash
+    :raise ValueError: if input array contains empty or null values
+    """
+    if not all(str_array):
+        # all values need to be non-null/non-empty
+        raise ValueError
+    str_array.sort()
+    blob = json.dumps(str_array, separators=(",", ":"), sort_keys=True).encode("ascii")
+    return sha512t24u(blob)
```
mcannon068nw, line 27: nice

server/src/metakb/transformers/identifiers.py
```diff
@@ -0,0 +1,103 @@
...
+    important that they are especially meaningful, but they should be consistent
+
+    :param source_prefix: prefix to use in ID namespace
+    :param combo_class: type of entity combination
+    :param operator: operator enum instance from concept set
+    :param ids: list of entity IDs to combine, must all be non-empty/non-null
+    :return: CURIE designating the combination in a deterministic way
+    :raise ValueError: if unrecognized combo_class type or if IDs list contains
+        null or empty values
+    """
```
mcannon068nw, line 87: does this care about order of the combination therapy, e.g. are

`["moa.therapy:trastuzumab", "moa.therapy:imatinib"]`

and

`["moa.therapy:imatinib", "moa.therapy:trastuzumab"]
`

treated as the same therapeutic group?

server/src/metakb/normalizers.py
```diff
@@ -254,7 +259,7 @@ def _normalize_concept(
         return normalizer_resp, normalized_id
 
 
-class NormalizerName(str, Enum):
+class NormalizerName(StrEnum):
```
mcannon068nw, line 262: Just learned about `StrEnum` from this. Seems like a neat little thing

mcannon068nw:
i left a review and a bunch of questions/comments. I think it looks good and will give approval tentative to a dev review

mcannon068nw (approved):

server/src/metakb/transformers/base.py
mcannon068nw: one more thing, does this address #178 anywhere?

server/src/metakb/repository/neo4j_models.py
jsstevenson: Ah thanks -- this should've been #665

server/src/metakb/transformers/base.py
jsstevenson: updated

server/src/metakb/transformers/base.py
jsstevenson: good question -- this shows up on the diff because it's moved around but my PR does not alter this method at all

server/src/metakb/transformers/base.py
jsstevenson: yep, that's exactly what this is for. ConditionSets can be arbitrarily deep; they don't appear at all in MOA (yet) but in CIViC there are two basic kinds, both of which occur relatively frequently

1) <Cancer> AND (<infant onset> OR <child onset> OR <teenage onset>) -- see eg https://civicdb.org/evidence/7365/summary

2) <Cancer> AND (<presenting phenotype A> AND <presenting phenotype B> AND ... ). This is VERY common in a bunch of the VHL assertions like https://civicdb.org/assertions/42/summary. 

I think we *need* to support #1 and we probably don't care about supporting #2 for a long time

jsstevenson: Added a note to the parent method about this

server/src/metakb/transformers/base.py
jsstevenson: I learned that it can also be a bunch of other things, but yeah, age of onset is gonna be what we're interested in

server/src/metakb/transformers/base.py
jsstevenson: Yeah made an issue about this here https://github.com/cancervariants/metakb/issues/726

I think that since we're getting stuff straight from CIViC, then we are already meeting the parameters defined by their workflow doc. But I would like to introduce a step where we can fine tune further if necessary down the line.

server/src/metakb/transformers/base.py
jsstevenson: Haha this was originally a match case block but I figured it was less showy to use `if isinstance`

server/src/metakb/transformers/catvars.py
jsstevenson: I think I wrote this and then realized we weren't capturing anything from CNVs. I can remove if we think it's dangerous to have the code here

server/src/metakb/transformers/civic.py
jsstevenson: Yeah idk. I think each source has to use source-specific judgments to make the leap from their own variant/profile models into a CatVar, so I was trying to think about shared logic that we could standardize against. In each source, you have to reimplement the "classifier", but if each source can accomplish this in a similar way, i think it'll be easier to maintain in the long run

server/src/metakb/transformers/civic.py
jsstevenson: shrug emoji

server/src/metakb/transformers/identifiers.py
jsstevenson: it should sort them a few lines down. I don't think there are known cases where source-provided name order is meaningful, but we need to sort them for the sake of deterministic hashing.

server/src/metakb/transformers/identifiers.py
jsstevenson: sha512t24ui hulud

server/src/metakb/transformers/catvars.py
mcannon068nw: if it doesn't break anything i don't think there's a problem with it staying. Maybe if anything, could add like a `TODO: revisit when CNVs` added somewhere around it or something just to not forget

katie-perry (approved):
Image

Issue #727, opened 2026-02-25
Programatically generate names for ProteinSequenceConsequence catvars


jsstevenson:
It's useful to have names (eg `"BRAF V600E"`) for catvars, but we don't really have a way to generate human-readable names for ProteinSequenceConsequence CVs. We can always grab them from sources, but that's a) a little icky, and b) leads to inconsistent names (e.g. we end up with a bunch of extra tags like `"(missense)"` in the search UI). We should use the normalizers/coolseqtool to get the gene name, and then generate the position change text.

### Use case

The below method will get used when constructing catvars (but dont worry about the rest of the catvar, the hard part is just generating the name for the allele)

### Acceptance criteria

A new method in the `metakb.transformers.base.Base` class that takes a VRS allele as input, and

1) validates that it is on a protein sequence,

2) figures out what gene it's associated with, 

3) returns a string of the form `"<gene name> <REF><POS><ALT>"`

for example

```
{
  "id": "ga4gh:VA.j4XnsLZcdzDIYa5pvvXM7t1wn9OITr0L",
  "type": "Allele",
  "location": {
    "id": "ga4gh:SL.t-3DrWALhgLdXHsupI-e-M00aL3HgK3y",
    "type": "SequenceLocation",
    "sequenceReference": {
      "type": "SequenceReference",
      "refgetAccession": "SQ.cQvw4UsHHRRlogxbWCB8W-mKD4AraM9y"
    },
    "start": 599,
    "end": 600,
    "sequence": "V"
  },
  "state": {
    "type": "LiteralSequenceExpression",
    "sequence": "E"
  }
}
```

becomes "BRAF V600E"

For now, we can throw `NotImplementedError` if

* it's anything more complicated than a SNP (ie if `len(state.sequence) != len(location.sequence`)
* `sequence` is unavailable in `location`

BUT both of those cases are theoretically solvable problems that we'll need to tackle eventually too

jsstevenson:
I'm calling this closed btw #735

PR #735, opened 2026-02-26
feat: add function for getting normalized name for protein sequence consequence catvars

katie-perry:
#727

### DIFF
--- a/server/src/metakb/transformers/base.py
+++ b/server/src/metakb/transformers/base.py
@@ -61,7 +61,6 @@ from metakb.normalizers import (
 )
 
 logger = logging.getLogger(__name__)
-NORMALIZED_VARIANT_NAME_EXT = "normalized_name"
 
 # Normalizer response type to attribute name
 NORMALIZER_INSTANCE_TO_ATTR = {
--- a/server/src/metakb/transformers/base.py
+++ b/server/src/metakb/transformers/base.py
@@ -61,6 +61,7 @@ from metakb.normalizers import (
 )
 
 logger = logging.getLogger(__name__)
+NORMALIZED_VARIANT_NAME_EXT = "normalized_name"
 
 # Normalizer response type to attribute name
 NORMALIZER_INSTANCE_TO_ATTR = {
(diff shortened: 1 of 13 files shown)


server/src/metakb/transformers/base.py
```diff
@@ -530,10 +531,98 @@ def _merge_mappings(
...
+
+        gene_name = None
+        transcript_mappings = self.vicc_normalizers.variation_normalizer.gnomad_vcf_to_protein_handler.mane_transcript.transcript_mappings
+        for protein_alias in protein_aliases:
+            accession = (
+                protein_alias.split(":", 1)[1]
+                if ":" in protein_alias
+                else protein_alias
+            )
+            gene_symbol = transcript_mappings.get_gene_symbol_from_refeq_protein(
```
katie-perry, line 606: nts: make ticket to fix typo in var norm

server/src/metakb/transformers/civic.py
```diff
@@ -669,6 +671,23 @@ def _get_psc_constraints(allele: Allele) -> list[Constraint]:
...
+                        annotated_variation_root
+                    )
+                except Exception:
+                    _logger.debug(
+                        "Unable to derive normalized_name for CIViC molecular profile %s",
+                        molecular_profile.id,
+                        exc_info=True,
+                    )
+                else:
+                    extensions.append(
```
katie-perry: I don't really know if I love putting this in the extensions/if that's what this is meant for but I'm not really sure where else to put it. I was struggling to find an existing pattern of where we have introduced variables that weren't in the VA-spec objects, so I put it in here for now.

server/src/metakb/transformers/base.py
```diff
@@ -530,10 +531,98 @@ def _merge_mappings(
...
+        ]
+        if not protein_aliases:
+            msg = (
+                f"Allele {allele.id} sequence reference {refget_accession} is not a "
+                "protein sequence."
+            )
+            raise ValueError(msg)
+
+        gene_name = None
+        transcript_mappings = self.vicc_normalizers.variation_normalizer.gnomad_vcf_to_protein_handler.mane_transcript.transcript_mappings
```
jsstevenson, line 599: dear god

server/src/metakb/transformers/civic.py
jsstevenson: the eventual goal is that it's `name` property in the CatVar object. so you can just set it 10 lines down from here

server/tests/unit/transformers/test_base_transformer.py
```diff
@@ -0,0 +1,135 @@
+"""Unit tests for Transformer base helpers."""
```
jsstevenson, line 1: 👍 nice

jsstevenson (dismissed):
👍 looks good to me, tests are great. Eventually this will get used to make the `name` property of the aggregate CatVar object, but what you have here should work for now

server/src/metakb/transformers/base.py
katie-perry: is this overkill? It might be overkill. 😬

jsstevenson (dismissed):

jsstevenson (approved):

PR #736, opened 2026-02-26
feat: add preflight checks for transform

katie-perry:
I was having some issues yesterday when I was trying to run `transform` and lost more time than I'd like to admit because I didn't realize I had some environment variables misconfigured (or forgot to set them at all). The transform will fully run, taking a few minutes, leading a dev to believe it's working properly, just to arrive at the end with an empty db and silent failures.

This code was debug code for myself, but I decided to open this PR in case it might be helpful to add to the cli. Everything _should_ work the same as before (I think. I didn't verify with the OG way of having the normalizers stood up on a local ddb instance) but with initial checks and probes before moving on. 

One thing I'm not super sure on is the pydantic version issue. It seems super niche to be including in the code like this, so I might end up taking that out. I'll comment on it to bring attention. 

I'm also unsure about exposing the errors - it seems like there's existing patterns in place to silence/log errors. When I was trying this out yesterday, though, I didn't see any errors logged to metakb.log when I was running the transform.

### DIFF
--- a/server/src/metakb/cli.py
+++ b/server/src/metakb/cli.py
@@ -3,7 +3,9 @@ to graph datastore.
 """
 
 import datetime
+import importlib.metadata as importlib_metadata
 import logging
+import os
 import re
 import tempfile
 from collections.abc import Generator
@@ -17,6 +19,7 @@ import boto3
 from boto3.exceptions import ResourceLoadException
 from botocore import UNSIGNED
 from botocore.config import Config
+from botocore.exceptions import ClientError, EndpointConnectionError
 
 from metakb import DATE_FMT, __version__
 from metakb.config import get_config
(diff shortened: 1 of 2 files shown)


server/src/metakb/cli.py
```diff
@@ -69,6 +73,150 @@ def _help_msg(msg: str = "") -> None:
...
+        ):
+            pydantic_version = importlib_metadata.version("pydantic")
+            variation_norm_version = importlib_metadata.version("variation-normalizer")
+            msg = (
+                "Variation normalizer failed due to a dependency compatibility issue.\n"
+                f"Detected pydantic=={pydantic_version}, variation-normalizer=={variation_norm_version}.\n"
+                "Known symptom: AttributeError involving `ValidationInfo` and "
+                "`warnings`.\n"
+                "Try pinning pydantic below 2.12 in this environment, e.g.:\n"
+                '  pip install "pydantic<2.12"'
```
katie-perry, line 197: this is the weird issue I ran into yesterday that I talked about in the description. I don't really like having this specific in here (I'll likely end up removing it) but left it in for now to call attention to this - should we be pinning a specific pydantic version between the metakb and normalizer repos? The difference in versions I had between metakb and variation normalizer was calling a failure. Matching my metakb pydantic version to the one in var-norm fixed it. 

I'm wondering if I should've looked more into it and created a discussion or a ticket...

jsstevenson: waaaaaa

Yeah if possible I'd like to see if we can just solve this. I'm guessing this will come up again in other places? As long as we have a new ticket to reproduce the error and then investigate its causes, I think this can stay the way it is

jsstevenson (approved):
👍 Yeah beside the pydantic version thing, I think these are all reasonable checks.

server/src/metakb/cli.py
katie-perry: Created a ticket https://github.com/cancervariants/metakb/issues/741

I'm making updates to another PR and having issues w the transform again so I'm going to merge this so I can see what's going wrong. I'll make sure to address this issue sooner rather than later

PR #758, opened 2026-03-10
feat: add get_all_ids and update_assertion_strength methods

### LINKED ISSUE #757: Create repository instance methods for getting all assertion IDs and updating assertion properties
I think this will be needed for the aggregation work, which will go something like

1) get all statement IDs
2) for each assertion ID,
3)  fetch the assertion from the DB
4)  recalculate its aggregate strength and star rating values
5)  if those values change from what's already in the DB, then
6)  update the assertion's properties in the DB

```
    def get_all_assertion_ids(self) -> list[str]:
        """Return all assertion IDs"""
        raise NotImplementedError

    def update_assertion(self, assertion: Statement) -> None:
        """Update assertion properties, e.g. aggregate strength or star rating

        Given an assertion, find the corresponding graph record of the same ID and
        update its properties

        :param assertion: new assertion object
        :raise ValueError: if ``assertion.id`` is ``None``
        """
        raise NotImplementedError
```

jsstevenson:
close #757 

Today on the drive home i was thinking about how I might need something like this to update aggregate strength values, so I sat down to see how it would work, and then that ended up being mostly enough. This adds methods to

* get all metakb assertion IDs (so you have a list of everything you need to check for updating)
* update strength (other values like a star rating should be even simpler than this)

### DIFF
--- a/server/src/metakb/repository/base.py
+++ b/server/src/metakb/repository/base.py
@@ -2,6 +2,7 @@
 
 import abc
 
+from ga4gh.core.models import MappableConcept
 from ga4gh.va_spec.aac_2017 import (
     VariantDiagnosticStudyStatement,
     VariantPrognosticStudyStatement,
@@ -106,3 +107,17 @@ class AbstractRepository(abc.ABC):
     @abc.abstractmethod
     def teardown_db(self) -> None:
         """Reset repository storage."""
+
+    @abc.abstractmethod
+    def get_all_assertion_ids(self) -> list[str]:
+        """Return all assertion IDs"""
+
+    @abc.abstractmethod
+    def update_assertion_strength(
+        self, assertion_id: str, strength: MappableConcept
+    ) -> None:
+        """Update strength associated with an assertion
+
+        :param assertion_id:
+        :param strength:
+        """
--- a/server/src/metakb/repository/base.py
+++ b/server/src/metakb/repository/base.py
@@ -118,6 +118,6 @@ class AbstractRepository(abc.ABC):
     ) -> None:
         """Update strength associated with an assertion
 
-        :param assertion_id:
-        :param strength:
+        :param assertion_id: ID of statement to update
+        :param strength: new strength concept
         """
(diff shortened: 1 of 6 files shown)


katie-perry (approved):
I didn't test this - might make more sense to test as we implement this, but the code looks good to me!

server/src/metakb/repository/base.py
```diff
@@ -106,3 +107,17 @@ def load_statement(self, statement: Statement) -> None:
...
+        """Return all assertion IDs"""
+
+    @abc.abstractmethod
+    def update_assertion_strength(
+        self, assertion_id: str, strength: MappableConcept
+    ) -> None:
+        """Update strength associated with an assertion
+
+        :param assertion_id:
+        :param strength:
```
katie-perry: These are extremely self explanatory but for the sake of consistency with other comments, should they have little descriptions? I'm not pressed about it tho (already approved)

server/src/metakb/repository/neo4j_repository.py
```diff
@@ -798,3 +798,26 @@ def teardown_db(self) -> None:
...
+            lambda tx: list(tx.run(queries_catalog.get_all_assertion_ids()))
+        )
+        return [r["s.id"] for r in result]
+
+    def update_assertion_strength(
+        self, assertion_id: str, strength: MappableConcept
+    ) -> None:
+        """Update strength associated with an assertion
+
+        :param assertion_id:
```
katie-perry: same here, again, not pressed. just asking in case we care about consistency. These are pretty self-documenting names lol

server/tests/unit/repository/test_neo4j_repository.py
katie-perry, line 1: nice

server/src/metakb/repository/base.py
jsstevenson: Ahhh I knew i was going to forget this

mcannon068nw (approved):

PR #764, opened 2026-03-11
feat: calculate aggregate evidence + strength of assertions

### LINKED ISSUE #734: Move evidence strength normalization to backend
## Background
To get a working version of MetaKB out by the deadline, I put this logic in the frontend, but we should do this on the backend. The current implementation uses this document for converting: https://docs.google.com/spreadsheets/d/1FpUmoXmDLVXsNgqog6A9q6o0jIOi_760j02rxIQGFMo/edit?gid=0#gid=0
This should remain the same for moving this code to the backend, so the above document is a good resource to refer to for this work

In the frontend, in `normalization.ts`, there is a function called `normalizeEvidenceLevelFromStrength` which takes a strength and maps it to a "normalized" strength. 
So for example, currently:
- CIViC assertions (AIDs) with strengths of "Level A" will be normalized to "A" (same for "Level B" -> "B")
- CIViC evidence (EIDs) remain the same
- MOA evidence is mapped from known terms to the corresponding VICC evidence level (ex: "authoritative evidence" will be normalized to "A")

## Work to be done
We should be calculating this value in the backend on transform and storing it in the Statement somewhere. We will need functions to handle normalizing the evidence strength for the different sources (since MOA uses different terms)

### Acceptance Criteria:
- add functions in the transformers for each ingested source (currently civic and moa but maybe more depending on when this ticket is worked on) to calculate the normalized evidence level according to the VICC Concept Vocabulary document linked above
- add `normalized_strength` property to Statement node (verify with devs where this should be? I'm not sure if we can extend the VA-spec `Statement` to add a new value like `normalized_strength` or if it could simply go in the `extensions`. Also `normalized_strength` is just a suggestion - feel free to name it something more descriptive)
- add unit tests to verify the new functions and test edge cases (discuss with Alex/stakeholders/other devs what to do if we can't determine a normalized strength)
- remove `normalizeEvidenceLevelFromStrength` from the frontend (also remove other deps such as `EvidenceCodeMapping`, `TermToEvidenceLevel` if unused)
- update logic in the frontend to properly display evidence level if needed (there may or may not be work to be done by this point. At the time of writing this ticket, evidence aggregation in the backend is still being worked on. In case the aggregated evidence rows do not have a normalized strength yet, implement that in this ticket and ensure the frontend ingests this correctly from the response)

This ticket is for calculating normalized evidence strengths on individual evidence items (as in, the supporting evidence items that make up an aggregate row (also referred to as an "evidence line")

### AC gherkins for specific cases:
(These should be helpful for writing unit tests!)

When I harvest and transform a record from MOA
And the evidence strength is "authoritative evidence"
Then the `normalized_strength` on the transformed record should be "A" (exact naming of the field can be up to the developer)

When I harvest and transform a record from MOA
And the evidence strength is "fda recognized evidence"
Then the `normalized_strength` on the transformed record should be "A" 

When I harvest and transform a record from MOA
And the evidence strength is "professional guideline evidence"
Then the `normalized_strength` on the transformed record should be "A" 

When I harvest and transform a record from MOA
And the evidence strength is "clinical evidence"
Then the `normalized_strength` on the transformed record should be "B" 

When I harvest and transform a record from MOA
And the evidence strength is "clinical cohort evidence"
Then the `normalized_strength` on the transformed record should be "B" 

When I harvest and transform a record from MOA
And the evidence strength is "clinical intervention evidence"
Then the `normalized_strength` on the transformed record should be "C" 

When I harvest and transform a record from MOA
And the evidence strength is "clinical observation evidence"
Then the `normalized_strength` on the transformed record should be "C" 

When I harvest and transform a record from MOA
And the evidence strength is "clinical case study evidence"
Then the `normalized_strength` on the transformed record should be "C" 

When I harvest and transform a record from MOA
And the evidence strength is "preclinical evidence"
Then the `normalized_strength` on the transformed record should be "D" 

When I harvest and transform a record from MOA
And the evidence strength is "inferential evidence"
Then the `normalized_strength` on the transformed record should be "E" <---- IMPORTANT: verify this with Alex. VICC evidence codes support a level E, but CIViC does not. Do we want to go along with their evidence knowledge model and drop the "E" or do we want to keep it for our MetaKB normalized strengths?

When I harvest and transform an evidence record from CIViC
And the evidence strength is one of "A", "B", "C", or "D"
Then the `normalized_strength` on the transformed record should be equivalent to its original value

When I harvest and transform an assertion record from CIViC
And the evidence strength is one of "Level A", "Level B", "Level C", or "Level D"
Then the `normalized_strength` on the transformed record should be "A", "B", "C", or "D" (without the "Level" prefix) (<--- side note: I'm not sure it's actually possible to have assertions with all of these values but it wouldn't hurt to have the logic handle all of them)

-------
(End helpful gherkins for unit tests. The following ones are for regression testing/making sure MetaKB still works as it did before this refactor)

Given I am on the MetaKB UI
And I perform a search
When I view the search results
And I click to expand an aggregated row of evidence on the result table
Then I should see each evidence item
And each supporting evidence item should have the normalized evidence strength displayed
And each supporting evidence item should have its original evidence strength displayed (as it was reported from the source)

Given I am a developer working on MetaKB
When I look through the code in `client`
Then I should NOT see any calculating of a normalized evidence level

### LINKED ISSUE #739: Finalize MetaKB assertions' `Method` definition and compute assertion strength/direction
If I'm not mistaken, higher-order MetaKB assertions need a `Method` of some kind:

https://va-spec.ga4gh.org/en/1.0/core-information-model/entities/information-entities/method.html#method

Currently I have some static text that needs filled in:

```python
METAKB_METHOD = Method(
    id="metakb.method:2026",
    name="MetaKB (2026)",
    reportedIn=Document(
        name="Wagnerds et al",
        title="MetaKB v2",
        doi="10.1038/1111-1-1111-111-1111",
        pmid="9999999",
    ),
)
```

We also need a method that takes a `list[EvidenceLine]` param and produces a strength value (I think I'm understanding this right?)

jsstevenson:
close #739 
close #734

* Add a function to take a set of evidence lines (i.e. from an assertion) and calculate what the aggregate strength and directionality values should be.
* Add code to update them on-the-fly when CDMs are being uploaded to the DB. Previously we had talked about leaving them as-is at this point, then writing a cleanup CLI command that would go through and recalculate all values; however, I realized that with the way the cypher queries are currently written, this would leave a whole big mess and I felt like it was easier to just update assertion properties at the same time that you add new evidence.
* Pull strength normalization logic out of the frontend, using a metakb display value extension attached to the GKS strength objects. 
* Move a lot of the assessment method related stuff to a new `methodology.py` module.
* Refactor some existing transformer code to make use of the assertion merger/recalculation stuff; this simplifies a somewhat wordy part of the old code

## Temporary decisions

* Basically doing strength calculation like so:
 
Screenshot 2026-03-16 at 1 04 57 PM

* Direction resolves to `NEUTRAL` in cases of strength ties (tie == any conflicting evidence of highest strength rank, regardless of relative quantity) 
* Tentatively, add the "display value" extension to source strength objects. I would rather put them on the VICC ev codes in the evidence line, but that would require some bigger frontend changes.

## Helpful notes?

* When it's time, I think a star rating calculation could just be added to the `aggregate_assertion_evidence()` function, and updating the property itself could happen in the `update_assertion_properties.cypher` query

## Stuff I'm still confused by

* How to do classification?


mcannon068nw:
> I'm not sure what system I should be using for the aggregate strength values. I think these are supposed to validate as AMP/ASCO/CAP level values, but then I'm not sure where the VICC evidence codes come from.

https://docs.google.com/spreadsheets/d/1FpUmoXmDLVXsNgqog6A9q6o0jIOi_760j02rxIQGFMo/edit?gid=0#gid=0

https://lucid.app/lucidchart/8b231bc8-f627-4997-9d9e-af59903ce348/edit?invitationId=inv_0887b2e5-b72c-4dd8-8e63-3d947d70ee4f&page=I_ORDsud6sfD#

mcannon068nw:
> How to handle cases where evidence lines of same strength have conflicting directions? This PR just defaults to neutral, but is there a tiebreaker?

I'm pretty sure Alex talked about this as a future to-do in a requirements gathering session. Maybe for now we just use the majority direction with some kind of flag in an extension somewhere?

jsstevenson:
> > I'm not sure what system I should be using for the aggregate strength values. I think these are supposed to validate as AMP/ASCO/CAP level values, but then I'm not sure where the VICC evidence codes come from.
> 
> https://docs.google.com/spreadsheets/d/1FpUmoXmDLVXsNgqog6A9q6o0jIOi_760j02rxIQGFMo/edit?gid=0#gid=0
> 
> https://lucid.app/lucidchart/8b231bc8-f627-4997-9d9e-af59903ce348/edit?invitationId=inv_0887b2e5-b72c-4dd8-8e63-3d947d70ee4f&page=I_ORDsud6sfD#

Screenshot 2026-03-11 at 9 01 55 AM

katie-perry (approved):

Issue #639, opened 2025-11-20
Calculate star rating confidence levels for aggregate evidence items


mcannon068nw:
### Feature description

We need to support methods to determine evidence quality for aggregate evidence according to our 4-star system as described previously:

https://github.com/cancervariants/metakb/discussions/549

### MetaKB Jr.
0-star: might not support
1-star: single submission from a clinical lab or online resource OR multiple, dissenting submissions
2-star: submissions from multiple evidence records that are concordant (CIViC AIDs demonstrate concordance)
3-star: submissions from ClinGen Somatic Cancer Variant Curation Expert Panels
4-star: knowledge from WHO / NCCN / FDA Pediatric Approvals / other regulatory or professional guidelines

### Use case

I'm always frustrated when I want to gain a sense of quality about my aggregate evidence but there is nothing to indicate a sense of trust or concordance.

### Acceptance Criteria

Methods must be in place that calculate evidence quality and make the result available in such a way for our front end results page to display a graphic of one to four stars

### Proposed solution

Need to implement two functions

1) Given a higher-order assertion and an incoming piece of evidence, put that evidence in the proper evidence line underneath the assertion

2) Given a higher-order assertion, calculate the correct star rating

For a first pass, we will implement this as a post-processing step; there should be a service function to loop through the DB and update every assertion star rating in-place

### Contribution

None

katie-perry:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Wagner-Lab-612eb99450b02c0012afa744/issues/gh/cancervariants/metakb/639?planning-poker) with Zenhub @jsstevenson @korikuzma

korikuzma:
> 3-star: submissions from ClinGen Somatic Cancer Variant Curation Expert Panels
4-star: knowledge from WHO / NCCN / FDA Pediatric Approvals / other regulatory or professional guidelines

We're going to have to do some work in civicpy to get this info

mcannon068nw:
Consider splitting out a ticket from this to civicpy to grab the information we need

mcannon068nw:
Some additional notes on the intent of the evidence star rating system:

The goal of this system is to provide a transparent signal of evidence quality and concordance for evidence aggregations in MetaKB Jr. 

[...]

PR #772, opened 2026-03-18
feat: add star rating to metakb assertions

katie-perry:
this takes care of #639 
This adds computation of a star rating given an assertion's evidence lines. It handles adding them to the extensions and updating the values if new evidence lines are merged in


server/src/metakb/services/manage_data.py
```diff
@@ -92,10 +92,15 @@ def _check_for_assertion_updates(
     db_assertion = repository.get_statement(assertion.id)
     if not db_assertion:
         return
-    if db_assertion.strength != assertion.strength:
-        # update assertion in place with new strength value
+    if db_assertion.hasEvidenceLines != assertion.hasEvidenceLines:
+        # update assertion in place with newly-aggregated evidence and derived fields
```
jsstevenson: pinning to think about this

jsstevenson: It might be practical to just remove this `if` statement and always update the assertion properties/strength. 

I think the `if` condition was written incorrectly, previously. I guess now what we could do is check whether merging our assertion into the DB assertion change's the DB assertion's strength/direction/extensions, but maybe that's getting too tricksy; loading data takes like 30 seconds so I think it's NBD to add a handful of extra queries.

server/src/metakb/transformers/methodology.py
```diff
@@ -225,6 +225,38 @@ class ViccConceptVocabEntry(BaseModel):
 # vicc concept ID -> vocab entry
 VICC_CODE_INDEX = {v.id: v for v in _vicc_concept_vocab}
 
+
+# --- Star rating classes
+class StarRatingReason(str, Enum):
```
jsstevenson, line 231: these can just be `StrEnum`

server/src/metakb/services/manage_data.py
katie-perry: oh yeah I did this bc any time the evidence lines change, we should recalculate the aggregate things. I guess this was to be more broad than like "if strength and direction and star rating (extensions) change, then merge them" but maybe being more specific here would be safer. I also need to think about whether this would actually work bc I forgot if this would actually compare the arrays or if it'll always be true bc the arrays are going to be at different addresses. I forgot how python does array equality

jsstevenson (changes requested):

server/src/metakb/transformers/methodology.py
```diff
@@ -309,6 +341,74 @@ def vicc_code_to_aac(
...
+            evidence_id = (evidence_item.id or "").lower()
+            evidence_strength = evidence_item.strength
+            strength_mappings = evidence_strength.mappings if evidence_strength else []
+
+            for mapping in strength_mappings or []:
+                mapped_code = mapping.coding.code
+                if getattr(mapped_code, "root", mapped_code) in {
+                    "e000001",
+                    "e000002",
+                    "e000003",
```
katie-perry: commenting for me - I need to revisit this logic bc of the changes from strength aggregation. The individual evidence items used to get assigned a VICC code and that's no longer the case, so this won't work for all cases

server/src/metakb/services/manage_data.py
jsstevenson: I think in practice this condition would always fire because you're effectively checking if the DB contains a record that you haven't uploaded yet

mcannon068nw: maybe naive question: does it matter for the calculation or db or anything if the before/after values for hasEvidenceLines are equivalent even if the evidence contributing these are different? (e.g. maybe in an update an evidence gets removed but another one is added of equivalent level so the end result is the same). Does that even ever happen?

katie-perry: I'm not sure if removing a piece of evidence and putting it somewhere else could ever happen, but I was considering the case where we add a piece of evidence that's the exact same as what we already have. We might be able to optimize it in the future to prevent some extra computation

jsstevenson (approved):

jsstevenson (approved):

PR #776, opened 2026-03-19
feat: add star rating column on frontend

### LINKED ISSUE #641: Implement confidence star rating visual on results table
### Feature description

Rows on our aggregate results table should display a column with a star rating visual as determine by our evidence confidence calculations.

Figma: https://www.figma.com/design/zrfq03lqd9tvoORTvHtj7Q/MetaKB-wireframe?node-id=0-1&p=f

### Use case

I am always frustrated when I want to get a sense of trust from where my evidence came from but there is no easy to understand indicator

### Acceptance Criteria

On the results page, there should be a star rating column that takes our confidence calculation results (#639 ) and displays the corresponding number of stars

https://www.figma.com/design/zrfq03lqd9tvoORTvHtj7Q/MetaKB-wireframe?node-id=0-1&p=f

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Implementation details

_No response_

### Potential Impact

_No response_

### Additional context

_No response_

### Contribution

None

katie-perry:
close #641


client/src/utils/results.ts
```diff
@@ -43,6 +43,12 @@ export interface NormalizedTherapy {
   therapyInteractionType: TherapyInteractionType
 }
 
+export interface StarRating {
+  starRating: number
+  // TODO: enum
```
katie-perry: I'll fix this

katie-perry: I removed this todo - I don't really know if it's worth having this as an enum on the frontend AND backend? that might cause issues in the future - we'd have to update in both places. I think it's safest to leave this as a string (otherwise if we update backend and forget to update frontend, the page would crash due to a typescript error)

jsstevenson: Yeah I think adding as a backend enum imported to the frontend by pydantic2ts sounds good. Doesn't need to happen right away

jsstevenson (approved):

PR #783, opened 2026-03-20
Study transformer 10 and base changes

hmcostello:
Transformer for nbl_target_2018_pub added, formatting changes in study transformer to align with other transformers. 

Base changes: disallow frequency generation for variants with no vrs_id (variants that weren't able to normalize). Changed frequency per cancer calculations to use oncotree code instead of cancer type. 

To run pl:
[run_cbp_pl.py](https://github.com/user-attachments/files/26150475/run_cbp_pl.py)
[run_cbp_pl.py](https://github.com/user-attachments/files/26150475/run_cbp_pl.py)

To test frequencies:
[validate_frequencies.py](https://github.com/user-attachments/files/26150463/validate_frequencies.py)
[validate_frequencies.py](https://github.com/user-attachments/files/26150463/validate_frequencies.py)


server/src/metakb/transformers/cbioportal/base.py
```diff
@@ -572,7 +572,7 @@ def _add_variant_frequency_columns(
         sample_col = "SAMPLE_ID" if "SAMPLE_ID" in combined.columns else None
         variant_col = "Gnomad_Notation"
         study_col = "STUDY_ID"
-        cancer_col = "CANCER_TYPE"
+        cancer_col = "ONCOTREE_CODE"
```
katie-perry, line 575: I am curious about these changes - since this is changed in the base, then does this mean that any of the other studies that use the old `CANCER_TYPE` column will no long work?

hmcostello: Hi Katie,

No, they all still work. This is only for constructing the tumor mutation recurrence rates. We compare variant per cancer type, and the oncotree code was more specific and standardized than the cancer_type, so I decided to change to using that instead.

mcannon068nw: reran transformer 9 under this branch and it still populated all the data fields. still works with the base change!

mcannon068nw (dismissed):
sorry this took so long! With the base change pointed out in the comments I wanted to try running older transformers to ensure that data was still being populated as we expected. Everything looks good!

mcannon068nw (approved):

Issue #649, opened 2025-11-21
Star rating histogram summary graphic


mcannon068nw:
### Feature description

In the summary infographics panel on our aggregate results page, we should display a histogram that summarizes the distribution of metakb-assessed star evidence ratings. This should take up approximately 1/3 of the panel.

### Use case

I'm always frustrated when I'm trying to assess available evidence quickly (I'm a busy clinical scientist) but there's no clear place to find evidence trust!

### Acceptance Criteria

There should be a histogram that summarizes the distribution of metakb-assessed star evidence ratings. This should take up approximately 1/3 of the panel and should in-take data based off the currently selected filter set

### Contribution

None

PR #786, opened 2026-03-26
feat: add star rating histogram + titles for graphs

katie-perry:
#649 

I also added a title for the variant/disease heatmap to make it more clear
Screenshot 2026-03-26 at 3 06 35 PM


mcannon068nw (approved):
very excited about this!!!

PR #789, opened 2026-03-27
feat: check for sc-vcep approved assertions for 3 stars

### LINKED ISSUE #762: Use SC-VCEP approvals to determine if an assertion is 3 stars
We have access to approvals in the transformer. We need to get a pkl file from staging which should have some dummy data we could use to wire this up for star ratings. 

The SC-VCEP information added in https://github.com/griffithlab/civicpy/pull/203 unfortunately will not help in this PR since we use raw CIViC records, and that PR implements the isApprovedVcep in the civic gks export classes/functions, which we are not currently leveraging in MetaKB

katie-perry:
close #762

### DIFF
--- a/server/src/metakb/transformers/civic.py
+++ b/server/src/metakb/transformers/civic.py
@@ -87,6 +87,17 @@ class MolecularProfileNameComponents:
 class CivicTransformer(Transformer):
     """A class for transforming CIViC to the common data model."""
 
+    @staticmethod
+    def _assertion_has_vcep_approval(item: civicpy.Assertion) -> bool:
+        """Check if any of the assertion's approvals is from a VCEP-approved org.
+        :param item: The civicpy Assertion to check
+        :return: True if an approval was found where the organization that approved it is an SC-VCEP
+        """
+        return any(
+            getattr(getattr(approval, "organization", None), "is_approved_vcep", False)
+            for approval in (getattr(item, "approvals", None) or [])
+        )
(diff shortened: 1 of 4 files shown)


server/src/metakb/transformers/civic.py
```diff
@@ -205,6 +216,16 @@ def _civic_claim_to_statement(
         elif isinstance(item, civicpy.Assertion):
             try:
                 statement = create_gks_record_from_assertion(item)
+                # TODO: Put this in civicpy instead.
```
katie-perry, line 219: nts: create a ticket for this

server/src/metakb/transformers/methodology.py
```diff
@@ -382,6 +382,12 @@ def calculate_star_rating(
     seen_directions: set[Direction] = set()
     evidence_count = 0
 
+    def _has_true_extension(statement: Statement, name: str) -> bool:
```
katie-perry, line 385: this might be nice in a util somewhere instead of a helper in this function but I wasn't sure where to put it...

jsstevenson (approved):

server/src/metakb/transformers/methodology.py
jsstevenson: I added something similar here https://github.com/ga4gh/vrs-python/pull/611/changes but I'm not sure if it's quite what you're looking for

Issue #792, opened 2026-03-27
Add a filter for star ratings


katie-perry:
### Feature description

It would be nice to let users filter by star rating 

### Use case

As a user, I would like to filter assertions by their star rating

### Acceptance Criteria

Given a user is searching MetaKB
Then they should see options to filter assertions by 1, 2, 3, and 4 stars
And when they click the filter for 1 star
Then they should see only assertions with 1 star

They should also be able to clear this filter
There should NOT be a search bar for this filter (that's only for filters with many, many options)
If they have multiple filters selected, it should show all assertions matching any of those selected (so if they have 1 star and 4 star selected, they should be able to see 1 star and 4 star assertions)

^ functionality subject to change depending on future iterations. At the last requirements gathering meeting, Alex said that individual evidence items should also have star ratings, and we'll have evidence line aggregates based on those. In the future, we should check with him how granular we want the filtering to go (should users be able to filter on just assertion star ratings? or also evidence item star ratings?)

### Contribution

None

jsstevenson:
done!

PR #793, opened 2026-03-27
feat: add filter for star ratings

katie-perry:
#792


mcannon068nw (approved):
do not have up to date image to display the ratings so could not poke around. The code itself looks fairly straight forward however.

PR #801, opened 2026-03-31
feat: latest evidence line structure + related fixes

### LINKED ISSUE #743: Free text protein switch queries (eg "BRAF V600E") are quite slow
### Describe the bug

For some reason, gene protein change queries are suddenly quite slow. This does not seem to be affecting other kinds of expressions, like HGVS queries.

### Steps to reproduce

Run a query like `https://pediatric.metakb.org/search?variation=IDH1%20R132H` on the production box (results seem to be cached so try to grab a new one off of CIViC)

### Expected behavior

should be <2s

### Current behavior

~10s response time

### Acceptance Criteria

Yeah uhhh it shouldn't be this slow. I am very frustrated when it's slow.

### Possible reason(s)

_No response_

### Suggested fix

_No response_

### Branch, commit, and/or version

asdfkhjads

### Screenshots

_No response_

### Environment details

prod

### Additional details

_No response_

### Contribution

None

### LINKED ISSUE #749: Update search HTTP endpoint, and frontend API consumer, with new aggregate ev structure
No need to manually group things anymore, this will happen in the DB

### LINKED ISSUE #752: Validate that nothing is missing after big refactor
* Are we still capturing all of the same civic/moa statements?
* Have any values changed?

### LINKED ISSUE #777: Sort by evidence level for evidence rows under aggregation
In an aggregation, the evidence should be sorted first by level (A at top, D at bottom, etc)

### LINKED ISSUE #780: Fix evidence structure per 2026-03-20 call
* Generally speaking, 1 evidence item gets 1 evidence line above it
* Aggregate evidence lines per tenets of star rating (e.g. there's a "2 star" evidence line collecting all 1 star evidence lines)
* Metakb assertion should use metakb method
* Star rating goes on the evidenceLine `evidenceOutcome` property, it gets a mappableconcept with the wagnerds et al system 
* Star ratings are currently stored on the aggregate. There should be evidence lines on single EIDs or AIDs that are rolled up to the aggregate. This will allow for detection of concordance/discordance across nested evidence lines all feeding up to the top level aggregation

### LINKED ISSUE #787: Assertions with only one evidence item are not getting proper star ratings
### Describe the bug

If an assertion only has one evidence item (as in, one evidence line and one evidence item. as in, only 1 evidence item total for the whole assertion), it seems to not get the proper star rating. On the frontend, I have it default to 1 star with the reason "Does not meet other criteria". This string is only present in the frontend, and any assertion on the frontend ends up getting as the reason, implying that the `star_rating_reason` is missing from the `extensions`. This also implies that `star_rating` is missing because with the current logic, one can't exist without the other

### Steps to reproduce

1. Transform MetaKB evidence
2. Notice star ratings are missing for assertions that only have one evidence item

### Expected behavior

The assertions should be assigned the correct star rating

### Current behavior

they are not receiving a star rating at all (I think)

### Acceptance Criteria

Given I am viewing an assertion
And that assertion only has one evidence item
Then that assertion should have a star rating

Given I am viewing an assertion
And that assertion only has one evidence item
And the evidence item is FDA-approved or a professional guideline 
Then that assertion should have a star rating of 4 (yes, even if there's only one evidence line. This criteria takes precedence over all else)

Given I am viewing an assertion
And that assertion only has one evidence item
And the evidence item is NOT FDA-approved or a professional guideline (so its normalized AMP/ASCO/CAP code is B, C, D, or E)
Then that assertion should have a star rating of 1 (I'm not sure if it's possible for a B, C, D, or E level to be more than 1 star - we may want to verify with Alex at some point. Maybe something could technically be level C if it's FDA-approved but outside indication? But then does it still count as FDA-approved?)

### Possible reason(s)

I'm a silly goose, and I call the calculate star rating in the merge_assertions function... merge_assertions does not get called when there's only one... 

### Suggested fix

_No response_

### Branch, commit, and/or version

`ev-aggregation-feature-branch`

### Screenshots

_No response_

### Environment details

not applicable here

### Additional details

_No response_

### Contribution

None

jsstevenson:
close #749,  close #780, close #743, close #787, close #777. And you know what? close #752 too

* handle new evidence structure/star rating stuff in transformers, repository class, API output, and frontend
* some misc cleanup as I came across it
* removed a lot of tests, frequently because they were components that were no longer being used. Need to write some new ones.

CI checks failing because a new Variation Normalizer release is needed, but if you manually install locally, things work.


client/src/components/ResultTable/ResultTableRow.tsx
```diff
@@ -2,12 +2,28 @@ import { useState, FC } from 'react'
...
+/* Dictate order that evidence appears underneath assertion **/
+const gradeOrder: Record<string, number> = {
+  A: 0,
+  B: 1,
+  C: 2,
+  D: 3,
+}
+const getGradeRank = (line: EvidenceLine): number => {
+  const grade = getEvidenceGrade(line.strengthOfEvidenceProvided)
+  return grade && grade in gradeOrder ? gradeOrder[grade] : 999
```
mcannon068nw, line 19: this `? x : y` pattern is pretty neat

mcannon068nw (approved):

client/src/components/ResultTable/ResultTableRow.tsx
jsstevenson: this is called a ternary expression -- 

in python the equivalent would be `return gradeOrder[grade] if grade and grade in gradeOrder else 999`

PR #803, opened 2026-04-03
Study transformer 17

hmcostello:
Study transformer for lgg_ctf_synodos_2025 added.

Several study transformers needed column name added to headers to import, for sequencing_source. 

Base changes:
Add mapping for new study
Fixed study transformers with incorrect sequence_source
Include both race AND ethnicity (Mapping for both to harmonize terms)

Harvester changes:
Ignore rows beginning with # for incoming dfs
Map two new studies

To test pl on this study:
[run_cbp_pl.py](https://github.com/user-attachments/files/26468205/run_cbp_pl.py)

To test pl on ALL studies (will take ~15 hrs):
[run_cbp_pl_all_studies.py](https://github.com/user-attachments/files/26468208/run_cbp_pl_all_studies.py)

To validate frequencies:
[validate_frequencies.py](https://github.com/user-attachments/files/26468211/validate_frequencies.py)

If you run validate_frequencies.py on the csv resulting from an all study run, you will see 99 mismatches in the "frequency_json_crosscheck_results.csv". This is not an issue in the data. The root cause is that multiple different Gnomad_Notation values can normalize to the same VRS ID - different representations of the same variant (e.g., across GRCh37 and GRCh38 studies) may resolve to the same canonical VRS ID.


mcannon068nw (approved):

mcannon068nw:
variant frequency numbers for both per study and across study look great! very interested to start plugging some of these in.

I will note that the gene_hgnc_id field does not appear to be populating with the bulk all studies run:

image

mcannon068nw (changes requested):
See if we can track down why the gene_hgnc_ids field is not populating as expected in the bulk run

mcannon068nw (dismissed):

mcannon068nw:
> See if we can track down why the gene_hgnc_ids field is not populating as expected in the bulk run

This ended up being user error on my part, i did not have my variant normalizer docker running during the bulk run

mcannon068nw (approved):

PR #807, opened 2026-04-08
Correct gnomad variants that can't normalize

hmcostello:
Some variants were not able to normalize, especially INSs and DELs, because the data used to construct the gnomad notation did not have a reference allele/tumor allele. It was present simply with a "-". 

I moved the position one base 5'-wise, and included that new reference base into the gnomad notation. 

So, from:
5-150404988-T-- to 5-150404987-GT-G


mcannon068nw:
go away stalebot

jsstevenson:
removing this from our active list but keeping it around for reference

PR #831, opened 2026-05-08
feat: add info page skeletons

jsstevenson:
This adds scaffolding/structure to fill in for about/info/etc pages (ie to hand off to our friend ChatGPT or copy off of old Alex grants).

The pages themselves are blank and it'd be relatively easy to reorganize as needed. Each info page lives in a section, and comes with a sidebar that lets you easily navigate to other pages in that section. 

Links to everything are also available on the footer.

## misc thoughts

* for the footer, I'd like to add some logos for respective organizations/institutions on the right-hand column, but I'm not sure who should/shouldn't be included (could just do VICC and call it a day)

Screenshot 2026-05-08 at 3 29 20 PM
Screenshot 2026-05-08 at 3 29 16 PM


mcannon068nw:
I don't think we should have the Contact Us button link directly to an issue creation template. It doesn't have to be exactly like dgidb, but I think we need a dedicate email address similar to help@dgidb

mcannon068nw (approved):

jsstevenson:
> I don't think we should have the Contact Us button link directly to an issue creation template. It doesn't have to be exactly like dgidb, but I think we need a dedicate email address similar to help@dgidb

yep -- would like to fill that in once an email address is available. Could also have a general "contact us" page listing other forms of reaching out based on kind of inquiry

PR #833, opened 2026-05-12
feat!: use async neo4j connection

### LINKED ISSUE #820: Use `async` for all neo4j queries

jsstevenson:
close #820 

basically, our routes have to run in an event loop because of the variation normalizer (because reasons)

but blocking (ie non-async) calls made in an event loop block the whole loop. this is fine if we receive a single request at a time, but a production metakb instance could get tripped up if multiple users were submitting queries at once. switching to `async` calls is laborious but saves us a lot of danger in that area moving forward


server/src/metakb/repository/queries/catalog.py
```diff
@@ -35,111 +36,112 @@ def _load_multiple_queries(filename: str) -> list[str]:
...
-    return _load("load_disease.cypher")
+def load_disease() -> LiteralString:
+    return cast("LiteralString", _load("load_disease.cypher"))
 
 
 @cache
-def load_phenotype() -> str:
-    return _load("load_phenotype.cypher")
+def load_phenotype() -> LiteralString:
+    return cast("LiteralString", _load("load_phenotype.cypher"))
```
mcannon068nw, line 87: I don't think I understand all these `cast()` calls. What is the difference between just typing out the string and coercing (is this the right word?) to a `LiteralString` type

server/src/metakb/restapi/meta.py
```diff
@@ -30,8 +30,8 @@ def service_info() -> ServiceInfo:
     "/stats",
     summary="Get basic statistics about MetaKB data.",
 )
-def stats(
+async def stats(
     repository: Annotated[AbstractRepository, Depends(get_repository)],
 ) -> RepositoryStats:
     """Provide stats for MetaKB data"""
-    return repository.get_stats()
+    return await repository.get_stats()
```
mcannon068nw, line 37: I get that the scope of this PR is to support async but I still always have trouble envisioning what `await` actually does in practice. What is the difference between `return repository.get_stats()` and `return await repository.get_stats()`? Or maybe I should ask what is the situation where the `await` version will still work properly but the non await will fail.

mcannon068nw (approved):

server/src/metakb/repository/queries/catalog.py
jsstevenson: 👍 yeah this really could've been done in another PR but it was bugging me

basically `cast()` makes the coercion explicit, it just helps type checkers, doesnt really do anything functional

server/src/metakb/restapi/meta.py
jsstevenson: we had a discussion in class

jsstevenson: to tie it back together, the `await` keyword specifically designates a place that a coroutine can resume. It's like a "pause point" -- the awaitable value is a coroutine/future, which means it's a thing that might not be done yet. the event loop circles around, says "is this done yet?" and if so, resumes at the `await`, but if not, the event loop finds something else to do for a minute before checking again.

PR #844, opened 2026-05-20
docs: update RTD build image (#796)

korikuzma:
readthedocs just sent out a notice that `ubuntu-20.04` is deprecated

~~Note: Will create a CAB ticket later~~ actually, this isn't touching any source code. So I think we don't need CAB ticket

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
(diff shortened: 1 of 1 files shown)


korikuzma:
@cancervariants/metakb-maintainers said we will have a release before June 1, so we can use the original change that exists in staging.

PR #868, opened 2026-07-01
fix: ensure up to date statement counts

jsstevenson:


server/src/metakb/repository/queries/get_counts.cypher
```diff
@@ -17,7 +17,7 @@ WHERE cv.id STARTS WITH "metakb"
 RETURN {label: 'num_variations', count: count(cv)} AS info
   UNION ALL
 MATCH (s:Statement)
-WHERE s.id STARTS WITH "civic" OR s.id STARTS WITH "moa"
+WHERE NOT (s.id STARTS WITH "metakb")
```
mcannon068nw, line 20: this change is because we will have more sources than civic or moa, thus it needs to be allow all statements rather than select those that start with specific source names.

I guess I am wondering why we are excluding specifically metakb starts. Are these the "top level" statements that are used to calculate aggregation stars or am i misremembering?

Do we ever expect any internal NCH data to start with a 'metakb' prefix (such as MCI or varcat)

mcannon068nw (approved):

server/src/metakb/repository/queries/get_counts.cypher
jsstevenson: yeah that's a good question. I wanted to change this because I realized that we'd forgotten to update it with previous new sources, and it seemed annoying to have to remember to do so for each new source in the future. I think this is a safe expectation... I've been prefixing MCI with `mci` and would probably want to do something similar for varcat

mcannon068nw (approved):

PR #872, opened 2026-07-08
feat: show evidence direction in client

### LINKED ISSUE #767: Display evidence directionality on front end
"direction" indicates whether the evidence supports or refutes the "worse outcome" or "better outcome", or is neutral

Could be as simple as a badge somewhere on the row

jsstevenson:
close #767 

Screenshot 2026-07-08 at 4 05 49 PM

### DIFF
--- a/client/src/components/ResultTable/EvidenceSummaryCell.tsx
+++ b/client/src/components/ResultTable/EvidenceSummaryCell.tsx
@@ -1,0 +1,102 @@
+import { Box, Tooltip, useTheme } from '@mui/material'
+import { Cell, Pie, PieChart } from 'recharts'
+
+import { EvidenceLevel } from '../../models/codings'
+import { getEvidenceGrade } from '../../utils/results'
+import { AssertionResult } from '../../utils'
+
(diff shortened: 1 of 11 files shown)


mcannon068nw:
will the options always be supports / refutes? 

I think some kind of badge or symbol would be snappier but this is good. Maybe bold + coloration of the word supports (blue) / refutes (red) could make it pop more but that also might be too loaded.

jsstevenson:
> will the options always be supports / refutes?

I don't remember if the schema binds this explicitly, but in practice it's always one of `supports`, `refutes`, or `neutral` (which we defer to if there is conflicting evidence underlying the assertion): https://va-spec.ga4gh.org/en/1.0/core-information-model/entities/information-entities/statement.html#statement

if it's cool w/ you, I'm gonna hold off on doing any explicit styling until we have everything we want on the page, then we can do a glow-up in one pass

mcannon068nw:
looks good! We can figure out specific styling at a later time

mcannon068nw (approved):

PR #874, opened 2026-07-09
feat: add snapshot command

### LINKED ISSUE #864: Add DB snapshot CLI command
+ corresponding repository methods

gist being: create a JSON dump of the DB in its present form that can be easily reloaded

jsstevenson:
close #864 

Add a command `metakb snapshot create` that dumps the DB to a JSON file which can be loaded with the existing `metakb load files` command

```shell
[ 864/feat-db-snapshot ⚙ .venv] code/metakb/server % metakb snapshot create --file here.json
Creating DB snapshot...
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████| 2713/2713 [00:41<00:00, 65.72it/s]
[ 864/feat-db-snapshot ⚙ .venv] code/metakb/server % head here.json 
{
  "assertions": [
    {
      "id": "metakb.assertion:--RA8AzIkeIg3RI_OS8biB1_T-HtEZMa",
      "type": "Statement",
      "extensions": [
        {
          "name": "metakb_star_rating",
          "value": {
            "primaryCoding": {
```

### DIFF
--- a/server/src/metakb/services/snapshot.py
+++ b/server/src/metakb/services/snapshot.py
@@ -1,0 +1,70 @@
+"""Provide tools for creating snapshots of stored data."""
+
+import logging
+from datetime import UTC, datetime
+from pathlib import Path
+
+import click
+from ga4gh.va_spec.base import Statement
+from pydantic import BaseModel
+from tqdm import tqdm
+
+from metakb import __version__ as metakb_version
+from metakb.repository.base import AbstractRepository
+
+_logger = logging.getLogger(__name__)
+
+
+class DbSnapshotMeta(BaseModel):
+    """Description of relevant snapshot metadata"""
+
+    name: str = "metakb_snapshot"
+    created_at: datetime = datetime.now(tz=UTC)
+    metakb_version: str = metakb_version
+
+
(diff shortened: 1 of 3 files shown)


server/src/metakb/services/snapshot.py
```diff
@@ -0,0 +1,70 @@
...
+
+class DbSnapshot(BaseModel):
+    """Data snapshot for backing up and reloading metakb data"""
+
+    assertions: list[Statement]
+    meta: DbSnapshotMeta
+
+
+class SnapshotError(RuntimeError):
+    """Raised when a database snapshot cannot be generated due to inconsistent data."""
```
mcannon068nw, line 34: Should this class have enums or mappings in it at all? What does an inconsistent database state and how would it arise

mcannon068nw (approved):

server/src/metakb/services/snapshot.py
jsstevenson: it's raised if an ID provided in the "get all IDs" query can't be retrieved by the "get statement by ID" query

so if somehow, like, half of a statement got deleted, then this could happen

PR #886, opened 2026-07-24
feat: add age of onset filter to frontend

jsstevenson:
Screenshot 2026-07-27 at 11 41 33 AM


client/src/components/SearchFilters/ChecklistFilter.tsx
```diff
@@ -12,9 +12,19 @@ import {
...
+export interface ChecklistFilterProps {
+  // title of filter section
+  title: string
+  // literal options to select from (eg disease names, star ratings, etc)
+  options: string[]
+  // subset of options indicating what's selected by the user
+  selected: string[]
+  // setter to set selections
+  setSelected: (values: string[]) => void
+}
```
mcannon068nw, line 25: Similar to the discussion for the significance values earlier, do we want the filters to be drawn based off what data is available or do we want consistence across searches (may be the wrong wording). Do we want the absence of data to be indicated by not having a filter or do we want a filter to exist that is clickable and will show no data when clicked

client/src/components/SearchFilters/StarRatingFilter.tsx
```diff
@@ -12,13 +12,21 @@ import {
...
-import type { FilterProps } from './types'
+
+export interface StarRatingFilterProps {
+  // literal options to select from (eg disease names, star ratings, etc)
+  options: string[]
+  // subset of options indicating what's selected by the user
+  selected: string[]
+  // setter to set selections
+  setSelected: (values: string[]) => void
+}
```
mcannon068nw, line 25: This made me think of a UX question for star rating selector behavior. Does the behavior of the star selector filter show at that level and below, or just that level? If I select four star does it show just four star or does it show four star + below.

client/src/components/SearchFilters/StarRatingFilter.tsx
```diff
@@ -12,13 +12,21 @@ import {
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
 import StarIcon from '@mui/icons-material/Star'
 import StarBorderIcon from '@mui/icons-material/StarBorder'
-import type { FilterProps } from './types'
+
+export interface StarRatingFilterProps {
+  // literal options to select from (eg disease names, star ratings, etc)
```
mcannon068nw: small thing, but are the XFilterProp comments copy pasted and do we want to make them more specific for each case

mcannon068nw (dismissed):

client/src/components/SearchFilters/StarRatingFilter.tsx
jsstevenson: yeah could make them more specific. originally this was a shared set of props but then I broke them out

client/src/components/SearchFilters/StarRatingFilter.tsx
jsstevenson: for the sake of underlying code simplicity, `options` here is the literal set of possible values (1_star, 2_star, etc) but behavior on the actual component is "at least" (eg clicking star #3 -> select only 3- and 4-star items). I would say this matches newegg-type shopping search behavior

client/src/components/SearchFilters/ChecklistFilter.tsx
jsstevenson: Currently we only show branches that correspond to actual options; ie, we won't show "young adult onset" if there are no young adult onset terms. I think it makes sense to hide these (analogy: you don't list every possible disease in the disease filter)

mcannon068nw (approved):

PR #891, opened 2026-07-29
Infra overhaul

jsstevenson:
* use setuptools-scm to set version
* clean up dependencies
* break beanstalk stuff into a specific dependency group
* perform deployment build in codebuild -- this is WIP


mcannon068nw:
@jsstevenson to follow up with @korikuzma on exact logistics for this

PR #899, opened 2026-09-11
feat: improved rating filter/other misc UI cleanup

### LINKED ISSUE #896: Add "& up" to star rating filter
See amazon search page for reference

jsstevenson:
close #896 

* Custom evidence level filter for A/B/C/D. I did *not* make this one perform "& up" functionality because I thought maybe there would be reasons to pick specific ev levels? Not sure if that's true. I think this could use an Alex review.
  * If we do need to do it as "& up" I probably need to redo the styling so it lines up better with star rating
* Add "& up" to star rating, remove popover. I think this is clearer
* Also make all filters closed by default
* Collapse "showing results for" typography into a caption underneath the entity name

Screenshot 2026-09-11 at 1 44 34 PM

I'm not 100% on how it looks centered but it's better than when it's left-justified 

I also played around with coloring to match the rating colors in the pie charts, but it ended up looking cartoonish


mcannon068nw:
What do we think the desired behavior for this should be? Currently it seems like clicking an evidence level filter only filters out aggregations that do not contain the desired evidence level. Is this what we want or should we be omitting evidence from aggregations that do not match with the clicked level, regardless of if they are inside an aggregation that contains the desired evidence level.

For example, clicking on Evidence Level B will still show me aggregations and evidence that map to level C or D. 

image

Do we want this or do we want it to remove individual evidence items from non-empty aggregations so that only B is left in the results table.

mcannon068nw (approved):

